PR #9, opened 2026-07-18
fix(grok): keep Auto-approve from flipping back to Safe

creomant:
## Problem
With Grok Build, selecting **Auto-approve** (always-approve) resets to **Safe** after a turn. Long-running tasks are then re-prompted for permissions if you forgot to re-adjust the mode every time.

## Cause
Several layers:
1. Permission-mode sync treated shared `full_access` as an unknown Claude SDK mode and mapped it to Safe.
2. Grok `applySelectedMode` only updated local state and did not call ACP `setMode` / mode config.
3. Launch `managed_config.toml` using the global `settings.permissionMode` instead of the Grok-projected mode, so Auto-approve on a Grok tab could still launch with `ask`.
4. Session create/load could push a default safe mode into the toolbar before the selected mode was applied.

## Fix
- Normalize shared + Claude-style permission mode values in the tab sync callback
- Refuse silent Auto-approve → Safe downgrades from session reports
- Apply selected Grok mode via ACP (`setMode`, with config-option fallback)
- Skip permission-toggle sync on session create/load; anti-clobber on mode reports
- Resolve launch `permission_mode` from Grok-projected provider settings
- Map native Grok mode ids (`always-approve`, `ask`, `default`, `bypassPermissions`)

## Tests
- Unit coverage for mode mapping, applySelectedMode, no-clobber, and tab sync guards
- Manual: set Auto-approve, run a multi-tool turn, toolbar stays Auto-approve; `.grimoire/grok/managed_config.toml` has `permission_mode = "always-approve"`

## Notes
Verified locally in Obsidian with Grok Build after ensuring only one Grimoire plugin folder was installed.


sandsaber:
Thank you very much for the careful investigation, implementation, and manual verification. Your work correctly identified the main causes: the Grok-projected permission was not used during launch, and the selected mode was not actually applied to the ACP session.

I incorporated that work into #11 and preserved your contribution with:

`Co-authored-by: Creomant <contact@creomant.com>`

I am closing this PR in favor of #11 because the final implementation needs to preserve a few additional architecture and security boundaries:

- Grok-native mode aliases stay inside the Grok adapter instead of being interpreted by the shared `Tab.ts` UI layer.
- Provider-reported session modes are treated as untrusted effective state and cannot overwrite or widen the persistent user-owned permission.
- The config-option fallback runs only for the structured JSON-RPC method-not-found error (`-32601`). Transport, authentication, validation, and policy errors are preserved without attempting a second mutation.
- The fallback uses the mode config ID discovered from the current session rather than assuming that it is always `mode`.

#11 also adds focused regression coverage for these boundaries and introduces `CONTRIBUTING.md` and a pull request template so that the repository expectations are clearer for future contributions.

This is a consolidation of the fix, not a rejection of your contribution. Your investigation and manual verification were genuinely valuable, and the replacement would not exist in its current form without your work. Thank you again.

Related issue #10 remains open until the behavior is validated separately with a live Grok session.

PR #32, opened 2026-08-04
refactor(chat): polish chat UI components, accessibility, and documentation

dreamofcoffee:
## Summary

Comprehensive UI/UX polish across chat components, accessibility, and Chinese documentation.

### 1. Conversation History Modal
- Removed the oversized "Conversation history" tooltip that covered the entire history modal.
- Replaced the broad accessible label with `aria-labelledby`, preserving accessibility while preventing the whole panel from triggering a tooltip.

### 2. Scroll-to-Bottom Interaction
- Removed the isolated scroll-to-bottom button beside the composer.
- Moved resume-auto-scroll and jump-to-bottom behavior to the bottom control in the translucent right-side navigation rail.
- Preserved scroll-state synchronization while removing the duplicate entry point and visual noise.

### 3. Tool Call Layout
- Changed tool-call width inside assistant responses to `calc(100% - 44px)`, reserving space for the translucent navigation controls and aligning the block more closely with response content.
- Fixed the completed checkmark shifting into the middle flexible column when an expanded Shell tool call hides its summary.
- Pinned result and status elements to their right-side grid columns so the status position remains stable in both collapsed and expanded states.
- Kept long tool output internally scrollable and wrapping instead of clipping horizontally.

### 4. Stop Response Button
- Restyled the control to match the bottom toolbar: 28px sizing, subtle surface, consistent radius, and matching hover behavior.
- Fixed a CSS specificity issue where the button's own `display: inline-flex` overrode its hidden state.
- The stop button is no longer shown for normal responses.
- It appears only when the current response detects subagent activity or a subagent lifecycle event.
- It hides automatically when the response completes, is cancelled, or is interrupted.
- The button can still cancel subagent responses; the existing keyboard interruption path remains available for normal responses.
- Removed the duplicate browser-native `title` tooltip and retained only the black Obsidian tooltip provided through `aria-label`.

### 5. Permission Dialog
- Redesigned the permission card hierarchy, header, command preview, borders, shadow, and responsive layout.
- Added a maximum height and internal scrolling for long commands.
- Removed duplicated command text such as "Execute: full command" when the command is already shown in the command block.
- Switched the dialog to `aria-labelledby` to prevent a tooltip from covering the entire card.
- Changed the three actions to a vertical stack of equal-width, equal-height rows.
- Removed their default borders, shadows, and filled backgrounds.
- Increased selector specificity to override Obsidian's native button styling reliably.
- Rows remain transparent by default and show a subtle small-radius background only on hover or keyboard focus.
- Reduced the radius to `4px`; the cancel action uses a soft red hover state.

### 6. Context Usage Tooltip
- Removed the large custom context tooltip and its second "tokens left" line.
- Replaced it with Obsidian's native plain tooltip.
- The tooltip now shows only "used / total context window", for example `29k / 258k tokens`.
- Changed tooltip placement from above the meter to below it.
- Preserved percentage display, the circular meter, and warning color above 80% usage.

### 7. Chinese README
- Revised `docs/readme/README.zh-CN.md` throughout, rather than limiting changes to the Google-related section.
- Aligned terminology with the actual settings UI, including provider, model selector, plan mode, reasoning effort, and usage indicator.
- Rewrote the Antigravity CLI and legacy Gemini CLI guidance.
- Added Antigravity limitations on Windows and its file-edit permission constraints.
- Fixed the relative link to `DISCLOSURES.md`.
- Updated the synchronized documentation version to `1.1.1`.

### 8. Verification and Deployment
- TypeScript type checking passed.
- ESLint passed.
- All targeted regression suites for this change set passed; later key combined runs covered 326, 203, and 133 tests respectively.
- Obsidian CSS review passed.
- Dependency review and release metadata validation passed.
- Isolated release-bundle loading passed.
- Opening `grimoire-view` from the release bundle passed.
- Release versions of `main.js`, `manifest.json`, and `styles.css` were copied into the active vault plugin directory.
- SHA256 hashes match between the release bundle and the installed plugin files.
- The existing vault `data.json` was preserved.


sandsaber:
@dreamofcoffee, I'm curious about the stop button. Why remove it for normal iteration?

sandsaber:
Thank you for all the work on this PR, and especially for sticking with it through several review rounds. There’s a lot of good work here. I tested the branch locally across providers and kept the main direction of the redesign.

I made a few follow-up adjustments on top of the latest version:

- I kept the new vertical permission actions, command preview, scrolling for long commands, keyboard behavior, and the `aria-labelledby` fix.
- I restored more of Grimoire’s visual identity in the permission header, including the violet accent and icons.
- I now show the full requested action or target instead of truncating it. I also preserved separate project and user permission choices. A permission dialog needs to make it obvious what the user is allowing, especially for persistent decisions.
- I removed the plugin version from the chat header. It’s still available in Settings, where it belongs.
- I kept the Stop button for turns where subagent activity is detected, and Esc still interrupts the active response. For now, this is a parent-turn stop. Provider background agents may continue according to their native lifecycle, so I’m not presenting it as individual agent cancellation.
- I’m still thinking about how the subagent UI could be better. Clearer grouping, more useful state rows, and possibly individual controls all make sense, but I’d rather handle that in a focused follow-up.
- TODO rendering now avoids showing the whole plan as completed while items are still unfinished. I still think the TODO presentation needs another pass, especially for providers that don’t publish every state transition.
- I fixed a few provider-specific issues found during local testing. Grok subagents now get normalized labels and states. Qwen child-agent streams no longer leak into the parent response, and its context-window indicator now reads the parent session correctly, including turns that use subagents.

These adjustments don’t replace the work you did. The permission hierarchy, vertical layout, long-command handling, keyboard controls, and accessibility changes are still the foundation.

Thanks again for putting so much time into this. No more changes are needed from you right now. I’ll do one final review of the complete diff for any unadvertised behavior before merging.

dreamofcoffee:
> @dreamofcoffee, I'm curious about the stop button. Why remove it for normal iteration?

In original ui, the stop button only appears when subagent is called, I compare with other plugins such as Claudian, they also do this. Well, if you like it behave like codex, you can change it to a single button with 2 state:send/stop according to the conversation status.

PR #34, opened 2026-08-05
Redesign the settings hub and harden provider discovery

dreamofcoffee:
## Problem

The settings experience had accumulated several related usability and reliability problems:

- one top-level tab per provider made navigation wide, visually heavy, and difficult to scan;
- provider setup, workspace resources, environment variables, and advanced controls were distributed across overlapping surfaces;
- provider cards tried to infer model availability from persisted discovery data instead of reporting the one reliable local signal: whether the CLI was present;
- Gemini, Qwen, Grok, OpenCode, MiMoCode, and Kimi Code only resolved manually configured binaries even though their runtimes support launching from `PATH`;
- changing the language from General did not reliably rebuild the mounted custom settings DOM, so newly opened tabs could remain stale or stop responding until the settings window was reopened.

Fixes: N/A — no linked issue.

## Root Cause

The settings shell was organized around provider-owned top-level tabs and eagerly mixed provider configuration with resource management. This made the tab bar scale with the provider registry and left no single place to manage shared skills, subagents, MCP servers, environment variables, and commands.

Provider status mixed persisted `discoveredModels` with CLI availability even though model catalogs can be synthetic, stale, provider-managed, or unavailable until a session starts. Several ACP CLI resolvers also validated only per-host and legacy configured paths, while the corresponding runtimes still fell back to a command name on `PATH`. The UI could therefore present model information that was not a reliable statement about the current installation while still falsely reporting that a PATH-installed CLI was missing.

Finally, the locale handler called `PluginSettingTab.update()`. With the declarative settings API, that did not reliably rebuild the currently mounted custom tab root and its lazy-render closures. Reopening settings created a fresh root, which is why the new language only became reliable after closing the settings window.

## Approach

- Replace provider-per-tab navigation with four stable top-level tabs: General, Providers, Advanced, and About.
- Render provider selection as a compact, responsive four-column card grid with an inline enable toggle and a full-width provider settings panel.
- Move the former General-page advanced controls into a direct, non-table General section inside the provider-neutral Advanced surface, alongside Skills, Subagents, MCP, Environment, and Commands.
- Aggregate provider-owned resources through `ProviderWorkspaceRegistry` services, keep runtime/built-in rows read-only, and route edit operations back through provider settings renderers.
- Add explicit edit/delete row actions and preserve confirmation before destructive writes.
- Remove the duplicate General environment entry and provider-page advanced sections.
- Lazily render each top-level tab once per settings render cycle, guard delayed renders with a generation token, and retain keyboard and horizontal-overflow behavior.
- Add provider-neutral executable lookup utilities for validated configured files, provider `PATH`, the enhanced GUI-process `PATH`, and provider-specific fallback locations.
- Use the shared lookup in Gemini, Qwen, Grok, OpenCode, MiMoCode, and Kimi Code resolvers; prefer Windows `PATHEXT` launchers such as `.cmd` over adjacent extensionless npm shims.
- Limit provider-card status to “CLI detected” or “CLI not detected”; do not start model-catalog discovery merely to populate the settings overview.
- Track the active declarative settings root and rebuild that same root after a locale change so all tab labels, lazy content, and event handlers immediately use the new locale.
- Add a dedicated settings hub stylesheet using Obsidian theme tokens for compact tabs, borderless provider cards, full-width details, responsive resource tables, and management modals.

## Architecture

Architecture notes:

The settings hub remains in `src/features/settings/` and consumes only provider-neutral registry contracts. It does not interpret provider-native conversation state or model identifiers. Provider-owned settings renderers, command catalogs, agent storage, MCP managers, model catalogs, and enablement remain registered through `ProviderRegistry` or `ProviderWorkspaceRegistry`.

Executable file lookup is a shared utility under `src/utils/` and is used by six provider-owned resolvers. Each provider still owns its command name, configured path projection, default install locations, cache, and reset behavior. OpenCode and MiMoCode were updated together, and the same launch-path behavior was checked across the other ACP providers.

The Advanced General section directly renders the existing settings controls rather than adapting them to the resource table. The other Advanced sections remain aggregation/presentation layers: mutations are delegated to the owning provider catalog or storage service, and unsupported or runtime-only resources remain read-only.

## Security And Privacy

Security notes:

CLI discovery adds filesystem checks but does not execute commands. Candidate paths are expanded, checked for existence, and required to be files before use. `PATH` lookup uses provider environment text plus the existing enhanced process path; it does not persist detected local paths or log them.

The provider overview no longer starts model-catalog refreshes for status display. Existing provider enablement still delegates any required initialization to the provider-owned model catalog; this PR does not broaden process launch or session behavior.

The Advanced environment list displays variable names and scopes, never values. Resource deletion remains an explicit user action behind the existing confirmation modal. Read-only resources do not expose deletion. MCP, agent, command, and environment writes continue through their existing owner storage APIs, followed by the existing reload/apply flows. No permission-mode, workspace-containment, credential, or external-path authorization behavior is widened.

## Verification

Automated commands run:

```text
npm test -- --runInBand tests/unit/features/settings/GrimoireSettings.test.ts tests/unit/providers/opencode/OpencodeCliResolver.test.ts tests/unit/providers/gemini/runtime/GeminiCliResolver.test.ts tests/unit/providers/grok/GrokCliResolver.test.ts tests/unit/providers/kimicode/KimicodeCliResolver.test.ts tests/unit/providers/mimocode/MimocodeCliResolver.test.ts tests/unit/providers/qwen/runtime/QwenCliResolver.test.ts tests/unit/style/settingsCss.test.ts
# 8 suites passed, 51 tests passed

npm run typecheck
# passed

npm run lint
# passed

npm run build:release
# passed review CSS, dependency, metadata, isolated bundle-load, and view-open checks

node --check dist/grimoire/main.js
# passed

npm test -- --runInBand
# attempted: 352 suites / 6,921 tests passed; 29 suites / 67 tests failed on existing
# Windows/POSIX path assumptions and zh-CN date-format expectations outside this change
```

Manual verification:

Not run in Obsidian after the final locale rebuild and Advanced/General migration changes. The release bundle was copied to a local vault plugin directory and `main.js`, `styles.css`, and `manifest.json` matched the release bundle by SHA-256. Before submission, exercise locale switching and all four tabs in Obsidian with at least one enabled PATH-resolved provider.

## Generated Artifacts

- [x] `package-lock.json` matches `package.json` when dependencies change. No dependencies changed.
- [x] Generated root bundles, `dist/grimoire`, and local-only artifacts are not included. `pr-text.md` is a local PR-description handoff and should not be committed with the source changes.

## Review Checklist

- [x] The PR is focused on one coherent problem: making the settings and provider-management surface scalable and reliable.
- [x] Protocol tests use real structured payloads and error codes. N/A — no protocol event shape changed.
- [x] Documentation and user-facing copy are in English. Localized Chinese labels are limited to the existing explicit locale path.


dreamofcoffee:
This change redesigns the settings pages, especially providers pages, like [Codian](https://github.com/bcs1037/codian). Well, there're still some details need to be polished, you can decide whether organize pages like this style.

mimakarov:
@dreamofcoffee gotcha! Let me look into it :) Great work! Thank you for your effort!  

I have like a design guidance, I would like to keep in one style overall. It is easy to fall for different element with different styles, and then it will be a mess, I believe.  I will finalize design stuff and as well put it here for references. 

Curios, how do you use Grimorie at your daily tasks?

dreamofcoffee:
> @dreamofcoffee gotcha! Let me look into it :) Great work! Thank you for your effort!
> 
> I have like a design guidance, I would like to keep in one style overall. It is easy to fall for different element with different styles, and then it will be a mess, I believe. I will finalize design stuff and as well put it here for references.
> 
> Curios, how do you use Grimorie at your daily tasks?

Something interesting is that I usually use the Claudian 😂，because it designs with elegant simplicity. But its biggest problem is OpenCode session lost after every Obsidian reboot. And its i18n is very mutilated.

Other plugins I refered when I redesign Grimoire are Codian & Claudian Plus. They give me an idea about to gather Grimoire settings in 4 tabs: general\providers\advanced settings\About.

image

I hope more features can be added to Grimoire, such as the support of provider Pi, external contexts manage(multi folders add and remove),

image

Well, there's a small problem is that model selector may be overflow when the chat tab is narrow, so that's why I try to reduce the width of this menu.

image

sandsaber:
@dreamofcoffee thanks for the feedback! Fair :) 
Do you have a custom theme in your Obisidian? I have default dark 

External context: Grimorie has it, but not folders, and yes, I agree with you that this feature should be improved.
изображение

Model selector, that is my pain, to be honest. The problem is that different providers have different naming styles. Like Qwen, open code. GLM, for example, you can use from the Qwen subscription, but as well, from claude code or opencode >_> 


Last time we did update it, and right now it is much more compact. But I have a few thoughts about it.

dreamofcoffee:
> Do you have a custom theme in your Obisidian? I have default dark

I use the Minimal theme, with light mode, a blue primary color #1259B5

sandsaber:
Pushed `70effd44` on top of this PR — the vault-level resource management batch for the redesigned settings hub.

**What's added**
- `VaultSkillCommandCatalog`: shared base for provider command catalogs with vault skill persistence. OpenCode, Grok Build, MiMoCode, and Kimi Code catalogs now build on it; Codex/Claude entries aligned via `storagePath` for cross-provider skill merging.
- Provider skill settings UI (`ProviderSkillSettings`) in the Qwen, Gemini, OpenCode, Grok, MiMoCode, and KimiCode settings tabs, plus Gemini and Qwen slash-command and subagent editors backed by new vault storage (`GeminiCommandCatalog` / `GeminiAgentStorage`, `QwenCommandCatalog` / `QwenAgentStorage`).
- Grimoire-managed MCP for ACP providers: per-provider `.grimoire/mcp/<provider>.json` storage (`AcpMcpStorage`) with an MCP settings manager in each ACP tab; enabled servers are injected into new/loaded ACP sessions and the process is recycled on changes (`reloadMcpServers`).
- The settings hub now renders from explicit per-provider `workspaceCapabilities` instead of hardcoded provider checks, and runtime command discovery is gated (`suppressAutomaticDiscovery`) so rendering settings never spawns CLI sessions.
- i18n hub copy across all 10 locales, README/docs/storage-boundary updates, and unit tests for the catalogs, storages, MCP runtime injection, and hub behavior.

**Also fixed**
- Case-only agent renames (e.g. `review` → `Review`) no longer fail with a misleading "already exists" error on case-insensitive file systems.

**Follow-ups noted during self-review** (non-blocking): an MCP server literally named `__proto__` would be dropped on save in `AcpMcpStorage`; saving a Gemini command whose existing TOML no longer parses should probably refuse instead of regenerating; string-form `tools` in Gemini agent frontmatter is not preserved on round-trip; some hub behaviors (semantic section extraction, `active-session-only` command discovery) could use dedicated tests.

PR #42, opened 2026-08-09
fix(build): revert TS 6/7 forks so the community scanner install resolves

sandsaber:
## Why

The Obsidian community scanner runs `npm ci --ignore-scripts` on the repo, and it failed on the `npm:` aliased TypeScript forks we added in #30 (`@typescript/native` @ `npm:typescript@^7.0.2` and `typescript` @ `npm:@typescript/typescript6@^6.0.2`). When that install fails, `obsidian` and `@types/node` are never resolved, so every import from those modules degrades to `any` and cascades into thousands of false `@typescript-eslint/no-unsafe-*` warnings on the scorecard — the maintainer flagged this in obsidianmd/eslint-plugin#182.

## What

Revert to plain `typescript@^5.8.3`, matching the official `obsidian-sample-plugin`:

- `package.json`: drop `@typescript/native`, set `typescript` to `^5.8.3`.
- `package-lock.json`: regenerated (drops the 20 TS7 platform-binary blocks + 3 alias blocks, −418 lines).
- `tsconfig.jest.json`: `ignoreDeprecations` `6.0` → `5.0` (valid under TS 5.x), and add `esModuleInterop` so default imports of `node:` builtins (`os`, `path`) used by the Antigravity provider and its tests resolve under ts-jest on TS 5.
- `ClaudePlanUsageStore.test.ts`: type the non-conforming rate-limit fixtures (string `resetsAt` values) as `Record<string, unknown>` so `tsc` on TS 5 is happy while still exercising the store's raw-input handling.

`^5.8.3` satisfies the `ts-jest` (`>=4.3 <7`) and `@typescript-eslint` (`>=4.8.4 <6.1.0`) peer ranges.

## Verification

- `npm ci --ignore-scripts` now passes in a clean dir on both **arm64** and **x64-linux** (the scanner platform), and `obsidian` + `@types/node` install correctly.
- `npm run typecheck` — green.
- `npm run test -- --selectProjects unit` — **6902 passed** (392 suites).
- `npm run lint`, `review:source`, `review:css`, `review:deps` — all green.

## Trade-off

This loses the faster TS 7 native compiler (`tsgo`) for `npm run typecheck`. The native compiler is referenced only as a devDependency (no CI/build/release path uses it), and typecheck on TS 5.9.3 is still fast here. If the speed matters, a follow-up can add it back as an opt-in `typecheck:fast` script that does not break the scanner install.

Ref: obsidianmd/eslint-plugin#182


sandsaber:
Closing in favor of #43, which also moves the type packages (obsidian, @types/node, typescript) into dependencies — empirically that is what drops the scanner's no-unsafe-* cascade from ~12,886 to 9. Reverting the TS forks alone (this PR) did not change the scorecard numbers.

PR #58, opened 2026-08-18
fix(antigravity): show vault skills in slash menu

jackycjrt:
## Summary\n- register a Vault skill catalog for Antigravity's slash menu\n- list .claude/skills and .agents/skills, including content-only SKILL.md files\n- expand selected Vault skills before passing the prompt to AGY\n\n## Verification\n- 52 focused unit tests passed\n- TypeScript and lint checks passed\n- production build completed


sandsaber:
Thanks @jackycjrt for the contribution — the vault skill menu works nicely as proposed, and expanding the invocation client-side was the right call for print mode.

I pushed a small follow-up commit on top of your branch to align a few details with repo conventions:

- the slash menu and the runtime now resolve skills through the same command catalog, so they always agree and skill frontmatter no longer leaks into the prompt
- the first-paragraph description fallback is now opt-in per root, so other providers keep their previous behavior
- appended note/editor context stays outside the skill-instruction wrapper
- empty skills and names that can't be typed as a `/` command are filtered from the menu

Core and runtime tests were added for all of the above.

Issue #59, opened 2026-08-19
Windows: unquoted prompt argument reaches cmd.exe when agy resolves to a non-.exe command


sandsaber:
## Description

On Windows, when the resolved `agy` command does not end in `.exe`, `buildAntigravityProcessLaunch` returns `shell: true` (`src/providers/antigravity/runtime/AntigravityProcessLaunch.ts:23-28`). Node's `spawn(command, args, { shell: true })` joins the command and arguments into a single `cmd /d /s /c` command line **without per-argument quoting**, so the prompt — the single argv element after `--print` (see `buildAntigravityPrintArgs` in `src/providers/antigravity/runtime/AntigravityChatRuntime.ts`) — is parsed by `cmd.exe`. Shell metacharacters in the text (`&`, `|`, `<`, `>`, `^`, `%VAR%`) can terminate the argument early, splice additional commands, or trigger variable expansion.

Since #58 the prompt can also contain expanded vault skill bodies, which widens the variety of text flowing through this path (same trust domain: the user's own vault).

On macOS/Linux the launch uses `['-lc', 'exec "$0" "$@"', command, ...args]` with `shell: false`, where arguments stay positional and are not re-evaluated — not affected.

## Steps to Reproduce

1. On Windows, configure Antigravity with a CLI path that resolves to anything not ending in `.exe` (a `.cmd`/`.bat` launcher or a bare command name).
2. Send a chat prompt containing cmd metacharacters, for example `summarize this & echo injected`.
3. Inspect the spawned command line: everything after the first unescaped metacharacter is interpreted by `cmd.exe`.

## Expected Behavior

The prompt reaches `agy` verbatim as a single argument on every platform; no part of user input or skill content is interpreted by a shell.

## Actual Behavior

On the Windows non-`.exe` fallback the prompt is concatenated into the `cmd.exe` command line unescaped, so metacharacters can alter or extend the executed command.

## Environment

- OS: Windows
- Affected code: `src/providers/antigravity/runtime/AntigravityProcessLaunch.ts:13-28` (pre-existing; re-surfaced while reviewing #58)

## Possible Fix Directions

- Resolve `.cmd`/`.bat` launchers explicitly and apply Windows argument quoting instead of relying on `shell: true`.
- Alternatively, pass the prompt via stdin or a temp file rather than argv on the Windows fallback path.

[...]

PR #68, opened 2026-08-20
fix(antigravity): pass vault to agy --print via probed --add-dir

### LINKED ISSUE #67: Antigravity times out on Windows when CLI path is a non-.exe wrapper; vault never added via --add-dir
### Summary
On Windows the Antigravity provider times out with no output. Any CLI path that doesn't end in `.exe` (a `.cmd`/`.bat` wrapper, or a PATH-resolved bare name) is launched with `shell:true` (via `cmd /c`), which mangles the multi-line `--print <prompt>` argument → `agy` gets broken args → no stdout → timeout. Separately, Grimoire never passes `--add-dir <vault>` to `agy --print`, so the agent's workspace doesn't include the vault; users use a `.cmd` wrapper as a workaround, which then triggers the `cmd /c` bug.

### Environment
- OS: Windows
- Grimoire: 1.0.x (Obsidian desktop)
- Antigravity CLI (`agy`): Windows build (reproduces on 1.1.x)
- Provider: Antigravity

### Repro
1. Set Antigravity CLI path to a `.cmd`/`.bat` wrapper that `cd`s to the vault and forwards args to `agy.exe` (common workaround to give the agent the vault directory).
2. New Antigravity chat → send any prompt.
3. Session hangs → times out, no response text.

### Expected / Actual
**Expected:** session responds in reasonable time; agent operates against the vault as its workspace (like other providers), so a wrapper is not needed.
**Actual:** no output; session waits and then times out. `agy` effectively receives broken/truncated args because invocation went through `cmd /c` (`shell:true`).

### Root cause (bundled `main.js`)
- Launch helper on Windows: path ending in `.exe` → `{command, args, shell:false}` (safe array spawn); any other path (`.cmd`, `.bat`, bare name) → `{command, args, shell:true}` → executed via `cmd /c`, which re-quotes and clobbers the multi-line `--print <prompt>` argument. `agy` gets malformed args → no stdout → Grimoire's stdout-recovery/timer path ends in timeout.
- Arg builder never prepends `--add-dir <vault>` for `--print`, so the agent's workspace never contains the vault root and it anchors on `~/.gemini/antigravity-cli/scratch` instead (same gap documented upstream, e.g. gsd-core#2176).

### Proposed fix
1. Launch non-`.exe` CLI on Windows with a safe array spawn (`shell:false`) where possible, so the multi-line `--print` argument round-trips intact.
2. Natively prepend `--add-dir <vault/cwd>` to the `agy --print` args array. Because `--add-dir` is repeatable and may not exist on older `agy`, capability-probe it (`agy --help | grep -- --add-dir`) before adding, mirroring how `--dangerously-bypass-hook-trust` is probed for other providers.

### Concrete code change (current bundled `main.js`)
The Antigravity argument builder is a minified function (token `TAe`) that collects the flag array. It is the only caller-owned builder for `agy --print` — other providers build args with their own functions, so the change below is scoped to Antigravity only.

**Current:**
```js
function TAe(t){let e=[];return t.permissionMode==="full_access"?...
```

**Patched** — inject `--add-dir <cwd>` as the first two elements:
```js
function TAe(t){let e=[];return t.cwd&&e.push("--add-dir",t.cwd),t.permissionMode==="full_access"?...
```

That is, right after `let e=[];return ` insert the guarded expression `t.cwd&&e.push("--add-dir",t.cwd),`. The `t.cwd &&` guard skips the flag when cwd is empty; `--add-dir` is repeatable and semantically valid for `agy`, so it only adds the vault to the agent workspace. Combined with pointing the CLI at `agy.exe` (→ `shell:false`), the multi-line prompt round-trips intact and the timeout disappears.

sandsaber:
## Summary
- Probe `agy --help` once per CLI command for `--add-dir` support (`AntigravityAddDirSupport`) and prepend `--add-dir <vault>` to `agy --print` args when advertised, so the agent workspace includes the vault root without wrapper scripts.
- The probe fails closed: spawn errors, timeouts, and the Windows empty-help quirk (agy 1.0.10 can exit 0 with empty stdout) launch without the flag, preserving current behavior on older CLIs.
- Concurrent queries share one in-flight probe; the result is cached per CLI command.

## Context
Fixes the `--add-dir` half of #67. The cmd.exe quoting half was already fixed for #59 in c119141 and ships in the next release.

## Test plan
- [x] Unit: probe detects `--add-dir` in help (stdout and stderr), fail-closes on spawn error/throw, caches per command, shares in-flight probes.
- [x] Unit: `buildAntigravityPrintArgs` prepends `--add-dir` first when requested and omits it when null.
- [x] Runtime: `query()` spawns the probe then the print run with `--add-dir <vault>`; omits the flag when help lacks it.
- [x] `npm run test -- --selectProjects unit`, `npm run typecheck`, `npm run lint`, `npm run build:release` all green.

Closes #67

### DIFF
--- a/src/providers/antigravity/AGENTS.md
+++ b/src/providers/antigravity/AGENTS.md
@@ -13,6 +13,7 @@
 - The AGY slash menu lists read-only vault skills from `.claude/skills` and `.agents/skills` through `VaultSkillCommandCatalog` (content-only SKILL.md files allowed). Because `agy --print` has no slash surface, `AntigravityChatRuntime` expands a leading `/skill-name` invocation client-side through the registered command catalog before launching AGY; keep menu filtering and the expansion grammar in sync.
 - Antigravity CLI 1.0.7 does not expose Gemini CLI's `--acp` flag; do not route it through `src/providers/acp/` unless a real ACP-compatible runtime is confirmed.
 - On Windows, resolved `.exe` commands launch directly; `.cmd`/`.bat` launchers and bare command names go through an explicit `cmd.exe /d /s /c` invocation with per-argument quoting in `AntigravityProcessLaunch` — never Node's `shell: true`, which concatenates the command line without quoting arguments (#59).
+- Grimoire probes `agy --help` once per CLI command for `--add-dir` support (`AntigravityAddDirSupport`) and prepends `--add-dir <vault>` to `agy --print` when advertised, so the agent workspace includes the vault root without wrapper scripts (#67). The probe fails closed: spawn errors, timeouts, or the Windows empty-help quirk launch without the flag.
 - Auxiliary workflows such as title generation, instruction refinement, and inline edit are unsupported until an Antigravity auxiliary runner exists.
 
 ## Boundaries
--- a/src/providers/antigravity/runtime/AntigravityAddDirSupport.ts
+++ b/src/providers/antigravity/runtime/AntigravityAddDirSupport.ts
@@ -1,0 +1,82 @@
+import { type ChildProcess, spawn } from 'node:child_process';
+
+import { buildAntigravityProcessLaunch } from './AntigravityProcessLaunch';
+
+const ADD_DIR_HELP_PROBE_TIMEOUT_MS = 10_000;
+
+const addDirSupportByCommand = new Map<string, Promise<boolean>>();
+
+/**
+ * Older `agy` builds predate `--add-dir`, and Windows builds can return exit
+ * code 0 with empty stdout (#67), so support is probed from `agy --help`
+ * instead of assumed. The result is cached per CLI command; callers that find
+ * no support simply launch without the flag, preserving existing behavior.
+ */
+export function probeAntigravityAddDirSupport(
+  command: string,
+  runtimeEnv: NodeJS.ProcessEnv,
+): Promise<boolean> {
+  const cached = addDirSupportByCommand.get(command);
+  if (cached) {
+    return cached;
+  }
+  const probe = detectAddDirSupport(command, runtimeEnv).catch(() => false);
+  addDirSupportByCommand.set(command, probe);
+  return probe;
+}
+
+export function resetAntigravityAddDirSupportCache(): void {
+  addDirSupportByCommand.clear();
+}
+
+async function detectAddDirSupport(
+  command: string,
+  runtimeEnv: NodeJS.ProcessEnv,
+): Promise<boolean> {
+  const launch = buildAntigravityProcessLaunch(command, ['--help'], runtimeEnv);
+  return new Promise<boolean>((resolve) => {
+    let child: ChildProcess;
+    try {
+      child = spawn(launch.command, launch.args, {
+        env: runtimeEnv,
+        shell: launch.shell,
+        stdio: ['ignore', 'pipe', 'pipe'],
+        windowsHide: true,
+      });
+    } catch {
+      resolve(false);
+      return;
+    }
+
+    let output = '';
+    let settled = false;
+    let timer: number | undefined;
+    const settle = (supported: boolean): void => {
+      if (settled) {
+        return;
+      }
+      settled = true;
+      if (timer) {
+        window.clearTimeout(timer);
+      }
+      try {
+        child.kill();
+      } catch {
+        // The child already exited between the event and this call.
+      }
+      resolve(supported);
+    };
+    const collect = (chunk: Buffer | string): void => {
+      output += chunk.toString();
+      if (output.includes('--add-dir')) {
+        settle(true);
+      }
+    };
+
+    child.stdout?.on('data', collect);
+    child.stderr?.on('data', collect);
+    child.on('error', () => settle(output.includes('--add-dir')));
+    child.on('close', () => settle(output.includes('--add-dir')));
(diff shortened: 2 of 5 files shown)


sandsaber:
## Multi-review summary

Four parallel review passes (logic/bugs, security, conventions, tests). Verdicts: **1 × request-changes, 3 × approve-with-comments**. Consolidated findings below; duplicated observations merged.

### Pre-merge

**1. Cancel during the probe window is silently dropped** *(bugs — major)*
`query()` awaits `probeAntigravityAddDirSupport` (up to 10 s) before spawning the print process, and `activeProcess` is still null in that window, so `cancel()` is a no-op. A stop click during the probe is lost entirely and the print run — with `--dangerously-skip-permissions` — still launches and runs up to `PRINT_TIMEOUT_MS`. Pre-PR, the gap between the status yield and `activeProcess` assignment was effectively synchronous, so this widens an uncancellable dead zone from ~zero to the probe duration. Same root cause affects `cleanup()` on plugin unload.

Suggested fix: a `cancelRequested` flag set in `cancel()`/`cleanup()`, checked after the probe resolves (bail out before `runPrint`), plus tracking the probe child so cancel can kill it instead of waiting out the timeout.

**2. Probe output accumulation is unbounded and listeners are never detached** *(security — minor, DoS-of-self)*
`collect` grows `output` without a cap (siblings use `appendLimited` at 64 KB), and after `settle` the `data` listeners stay attached. On the Windows `cmdShell` path, killing `cmd.exe` orphans the grandchild `agy`, which keeps writing into `output` after the 10 s timeout — memory growth is not bounded by the timer in that scenario. Fix: cap accumulation (fail-closed past the cap) and `destroy()` both streams in `settle`.

**3. Substring match is not boundary-anchored** *(bugs/security/tests — minor)*
`output.includes('--add-dir')` matches a hypothetical `--add-directory`-style flag; a false positive poisons the permanent cache (every launch fails with an unknown flag until reload). `--no-add-dir` does **not** match — verified. Tighten to `/--add-dir\b/` and pin both cases with tests.

**4. Vault path lands in debug logs verbatim on Linux** *(security — minor)*
`redactSensitiveText` covers `/Users`, `/tmp`, `/Volumes`, drive letters, etc., but not `/home`, `/root`, `/mnt`, …, so `print.spawn` `argsSummary` now embeds the absolute vault path on Linux, against the "no paths in debug logs" rule. Pre-existing `command` field shares the gap; the cheap airtight fix is call-site masking in `summarizeAntigravityPrintArgs` (`<vault-path>`, mirroring the existing `<prompt>` rewrite).

**5. Missing tests for two stated contracts** *(tests — important)*
- The 10 s timeout path has zero coverage; house precedent exists (`AcpSubprocess.test.ts` fake timers, `advanceTimersByTime`). The test should also pin that late help output after the timeout cannot flip the cached result.
- Negative-result caching is unpinned (only `true` is cached today); a regression to success-only caching would re-spawn `--help` every turn and stay green.

### Nice-to-have

- **Probe failure instrumentation** *(conventions)*: `supported: false` is indistinguishable between timeout, empty help, and spawn error — exactly the Windows quirk class this probe exists for. Sibling spawn modules log `*.timeout`/`*.spawnError` at `error` level; the probe should too.
- **Windows tree-kill** *(bugs — minor)*: killing the `cmd.exe` wrapper orphans `agy --help`; bounded impact (help exits on its own), mirrors the pre-existing `runPrint` kill pattern.

### Verified clean

- Event-loop correctness of the probe (listener attachment, settle-once, timer cleanup, sync-throw fail-close) — sound.
- `shell: false` on every launch branch; vault path is a discrete argv element; `quoteWindowsArgument` implements canonical MSVCRT quoting (`%` limitation pre-existing and documented from #59).
- Probe does not expand the trust boundary of the already-trusted CLI binary.
- Layering (probe confined to provider dir, no core/features leakage), comment policy, no `console.*`, test placement/style, provider `AGENTS.md` bullet accuracy — all pass.
- Existing tests unaffected by the extra probe spawn (transcript/multibyte tests call `runPrint` directly); module isolation makes the per-file cache reset correct and load-bearing; `setImmediate` barriers are deterministic.

sandsaber:
Review findings addressed in 55ff6d0:

1. **Cancel during probe** — the runtime now tracks the probe child (killed by `cancel()`/`cleanup()`) and checks a `cancelRequested` flag after the probe resolves, ending the turn with `done` instead of launching `agy --print --dangerously-skip-permissions` after a stop click. Covered by `drops the print run when cancelled while the add-dir probe is in flight`.
2. **Unbounded output / detached listeners** — probe accumulation is capped at 64 KB (fail-closed past the cap, mirroring `appendLimited`), and `settle` destroys both pipes so an orphaned writer under a killed `cmd.exe` wrapper gets EPIPE instead of streaming into memory. Covered by the output-cap test.
3. **Substring anchoring** — now `/--add-dir\b/`; `--add-directory` and `--no-add-dir` both pinned to `false`.
4. **Vault path in debug logs** — `print.spawn` `argsSummary` masks it as `<vault-path>` (same pattern as `<prompt>`), asserted through the real `recordDebugLog` call.
5. **Missing test contracts** — timeout path (fake timers, incl. late-output-after-timeout cannot flip the cached result) and negative-result caching are now pinned.

Full local gate green: 7032 unit tests, typecheck, lint, `build:release`.

PR #71, opened 2026-08-21
fix(antigravity): stdin stream-json transport and long-turn timeouts (#69, #70)

### LINKED ISSUE #69: Antigravity: spawn ENAMETOOLONG on Windows once the conversation grows (whole transcript is passed as one argv argument)
### Summary

On Windows the Antigravity provider fails with `spawn ENAMETOOLONG` as soon as a conversation gets long. `buildAntigravityPrintArgs` ends with `args.push('--print', spec.prompt)`, and `spec.prompt` is not just the user message: `buildAntigravityPrintPrompt` concatenates the last 12 chat messages plus `<current_note>`. That whole transcript therefore travels as a single argv argument, and `CreateProcess` rejects command lines longer than ~32767 characters. `shell: false` does not help — this is the OS argument-vector limit, not a quoting problem, so it is a different failure from #59.

The failure is not an edge case: it is guaranteed once the transcript exceeds the limit, and every later turn in that tab fails the same way.

### Environment

- OS: Windows 11
- Grimoire: 1.1.7 (Obsidian desktop)
- Antigravity CLI (`agy`): Windows build, 2026-08-20
- Provider: Antigravity, `permissionMode: full_access`

### Repro

1. Open an Antigravity chat and work until the running transcript passes ~32 KB (roughly a dozen substantial turns, or fewer if a large `<current_note>` is attached).
2. Send one more message.
3. The turn fails immediately with `spawn ENAMETOOLONG`; every subsequent turn in the tab fails too, because the history only grows.

### Expected / Actual

**Expected:** the prompt reaches `agy` regardless of transcript size, as it does for providers that stream input over stdin.
**Actual:** `spawn` throws `ENAMETOOLONG` before the process starts, so there is no stdout, no log file, and no diagnostic beyond the errno.

### Evidence

Direct measurement of the two channels on the same machine, same Node runtime that Obsidian ships (v22):

```js
// 200 KB through stdin
const child = spawn('python', ['-c', "import sys,json;print(len(json.loads(sys.stdin.read())['message']['content']))"],
                    { stdio: ['pipe', 'pipe', 'pipe'], windowsHide: true });
child.stdin.write(payload200kb); child.stdin.end();
// -> exit 0, child received 200000 characters

// the same 200 KB through argv
spawn('python', ['-c', 'pass', 'y'.repeat(200000)], { stdio: 'ignore', windowsHide: true });
// -> throws synchronously: Error: spawn ENAMETOOLONG (errno -4064)
```

### Root cause

`src/providers/antigravity/runtime/AntigravityChatRuntime.ts`:

- `buildAntigravityPrintArgs()` → `args.push('--print', spec.prompt)`;
- `buildAntigravityPrintPrompt()` / `buildAntigravityPromptText()` build `spec.prompt` from the last 12 messages plus the current-note context;
- `runPrint()` spawns with `stdio: ['ignore', 'pipe', 'pipe']`, so stdin is unavailable as an input channel.

### Proposed fix

`agy` already supports NDJSON over stdin, which has no length limit:

```
agy --add-dir <vault> --dangerously-skip-permissions --log-file <path> \
    --input-format stream-json --output-format stream-json
stdin: {"event":"user","message":{"role":"user","content":"<prompt>"}}\n
```

Concretely:

1. In `buildAntigravityPrintArgs`, replace `--print <prompt>` with `--input-format stream-json --output-format stream-json`. Note that `agy` rejects the combination of `--print` and `--input-format stream-json` explicitly, so the prompt must move to stdin.
2. In `runPrint`, open stdin (`stdio: ['pipe', 'pipe', 'pipe']`), write one NDJSON line, then `end()`. A `stdin.on('error')` handler is required: if the child exits early (for example on an empty prompt, where `agy` exits 1), the EPIPE arrives as a stream `error` event and would otherwise become an uncaught exception in the renderer.
3. Parse stdout as NDJSON and take `result.response` from the last `{"event":"result"}` line; surface `status:"ERROR"` with `result.error` instead of an empty string. The transcript fallback still works: `agy` keeps writing `--log-file` in this mode, so `extractAntigravityConversationId` / `extractAntigravityAppDataDir` continue to resolve.
4. Watch the stdout accumulator: with a fixed tail cap (`appendLimited`, 64 000 chars in the current code) a `result` line longer than the cap loses its head and stops parsing. Either raise the cap or parse line-by-line as the SDK transport does.

The attached `agy-stream-json-events.ndjson` documents the full event vocabulary observed: `init`, then `step_update` frames whose payload lives under `step_update` (`step_type` ∈ `user_input`, `checkpoint`, `agent_response`, `tool`; `state` ∈ `ACTIVE`, `DONE`), then a final `result`. Text arrives as `text_delta` chunks on `ACTIVE` `agent_response` frames plus a remainder on `DONE`; the concatenation equals `result.response` exactly (measured: 934 + 2356 = 3290 characters), so the switch also makes token streaming possible without duplicating the final answer.

### Notes

- Verified locally by patching the bundled `main.js` of 1.1.7 along these lines: nine consecutive real sessions, no `ENAMETOOLONG`, including turns of 10 and 11.6 minutes.
- `--add-dir` is unaffected and remains compatible with `stream-json` (`init.cwd` picks up the vault); the probe added in #68 keeps working.
- Happy to send a focused PR for this if you would like it.

[grimoire-antigravity-evidence.zip](https://github.com/user-attachments/files/31272941/grimoire-antigravity-evidence.zip)

### LINKED ISSUE #70: Antigravity: long turns die after 5 minutes — agy's default --print-timeout plus Grimoire's absolute PRINT_TIMEOUT_MS
### Summary

Antigravity turns that legitimately need more than five minutes are killed twice over. `agy` has its own `--print-timeout` (default `5m0s`) which Grimoire never sets, and Grimoire additionally applies `PRINT_TIMEOUT_MS = 5 * 60 * 1000` as an absolute limit on the whole request rather than as an inactivity limit. Whichever fires first, the user loses the turn and sees no usable reason.

### Environment

- OS: Windows 11
- Grimoire: 1.1.7 (Obsidian desktop)
- Antigravity CLI (`agy`): Windows build, 2026-08-20
- Provider: Antigravity, model `Gemini 3.7 Flash (Low)`

### Repro

1. Ask the Antigravity agent for work that takes more than five minutes — several tool calls over a large workspace is enough.
2. Watch the turn die at the five-minute mark.

### Expected / Actual

**Expected:** a turn that keeps producing output keeps running; only a genuinely stalled process is killed.
**Actual:** the run is aborted at five minutes. The user-visible message is `Antigravity CLI exited (code 1)`, which hides the structured reason `agy` did emit.

### Evidence

From `agy`'s own log (attached as `agy-print-mode-timeout.log`):

```
I0820 18:41:16.913146 printmode.go:173] Print mode: starting (promptLength=0, model="Gemini 3.7 Flash (Low)", conversationID="")
...
E0820 18:46:22.418942 printmode.go:521] Print mode: timed out after 1498 polls (printed=21)
I0820 18:46:22.418942 manager.go:739] CLI store manager shutting down
```

Wall clock: 18:41:16 → 18:46:22, i.e. the 5-minute default. The matching final frame (attached as `agy-print-timeout-result.json`):

```json
{"event":"result","result":{"status":"ERROR","response":"","error":"timeout waiting for response",
 "duration_seconds":227.265714,"num_turns":1,
 "usage":{"input_tokens":62813,"output_tokens":1033,"cache_read_tokens":264844,"total_tokens":63846}}}
```

Worth flagging: `duration_seconds` is **not** wall-clock time for the run — it reads 227 s for a run that took 306 s. Diagnosing this from `result` alone is misleading; the log line is the reliable signal.

At the moment of the abort the agent had completed nine foreground `run_command` calls and was in the middle of the tenth, so this was a healthy, working session — not a hang.

### Proposed fix

1. Pass `--print-timeout` explicitly from `buildAntigravityPrintArgs` instead of inheriting the 5-minute default, and keep it just below Grimoire's own ceiling so `agy` reports a structured `result` rather than being killed from outside.
2. Turn `PRINT_TIMEOUT_MS` into an **inactivity** timer: reset it on every stdout/stderr chunk and keep a separate absolute ceiling. In `stream-json` mode each tool transition emits a frame, so an active run refreshes the timer naturally while a silent hang still dies quickly.
3. When the process exits non-zero, parse the NDJSON buffer before falling back to the generic message: `result.error` (`"timeout waiting for response"` here) is far more actionable than `exited (code 1)`. Preserving the `--log-file` contents on failure instead of unlinking it unconditionally would also help, since that log is the only place where the real wall-clock cause appears.

### Notes

- Verified locally on 1.1.7 with `--print-timeout 29m` plus an inactivity timer of 300 s and a 30-minute ceiling: turns of 10 and 11.6 minutes now complete, and stalled sessions are still cut in five minutes instead of thirty.
- Depends on the stdin transport from the #69  issue only for point 3 (there is no NDJSON to parse without it); points 1 and 2 are independent.

[grimoire-antigravity-evidence.zip](https://github.com/user-attachments/files/31273034/grimoire-antigravity-evidence.zip)

sandsaber:
## Summary

Fixes #69 and fixes #70 for the Antigravity provider.

**#69 — `spawn ENAMETOOLONG` on Windows.** The whole conversation transcript travelled as a single argv argument to `agy --print`, so past ~32k characters `CreateProcess` rejected the command line and every later turn in the tab kept failing. When `agy --help` advertises support (one probe now covers `--add-dir`, `--input-format`/`--output-format`, and `--print-timeout` via the new `AntigravityCliCapabilities`), the run launches with `--input-format stream-json --output-format stream-json`, the prompt is written as a single NDJSON user event to stdin (EPIPE from an early-exiting child is handled), and stdout is parsed line-by-line by `AntigravityStreamJson` so a `result` frame longer than any fixed tail cap still resolves. `status:"ERROR"` frames surface their structured `result.error` instead of a bare `exited (code 1)`. CLIs without `--input-format` keep the legacy `--print` argv transport.

**#70 — healthy long turns killed at five minutes.** The single absolute `PRINT_TIMEOUT_MS` is replaced by a 5-minute inactivity timer refreshed by every stdout/stderr chunk plus a 30-minute absolute ceiling, and `--print-timeout 29m` is passed when supported so `agy` self-terminates with a structured result frame before Grimoire has to kill it. The `--log-file` is preserved after failed runs (removed on success and on routine user cancels) because it is the only place `agy` records the real wall-clock cause.

## Review hardening (three multi-agent review rounds)

- Settle on `close` instead of `exit`: Node can emit `exit` before the final stdout chunks drain (~46% of fast writers, measured), which lost the result frame.
- Only conclusive `--help` probes are cached — cancelled (including the Windows `(1, null)` kill shape), timed-out, and externally-signaled probes are retried next turn instead of pinning the session to the legacy argv transport.
- A 2-second drain grace forces the stdio streams closed after `exit` if `close` lags (orphaned `cmd.exe` wrapper holding the pipes), so Stop still settles.
- Per-run tokens keep a draining old turn from clearing a newer turn's process handles; user cancels are tracked per killed child process.
- A timeout that arrives after a complete result frame resolves with that answer instead of discarding it.

## Testing

- 112 antigravity unit tests (transport, parser incl. CRLF/multibyte/oversized frames, probe caching semantics, dual timers under fake timers, cancel paths, log retention, a FIFO-based drain-race regression test).
- Full local gate green: 7070 unit + 216 integration tests, `typecheck`, `lint`, `build:release` incl. Obsidian community-review gates.

## Deferred / to verify on a real Windows machine

- SIGTERM targets the `cmd.exe` wrapper on the default launch path; killing the whole process tree (`taskkill /T` equivalent) may be needed if orphaned `agy` processes are observed after cancel/timeout.
- The capability-probe cache is keyed per CLI command only (not env), so an in-place CLI upgrade keeps stale capabilities until restart.


greenkms:
The shape of #71 matches what I ended up with locally, including the details that only show up once you run it (settling on `close` rather than `exit`, and line-by-line stdout parsing instead of a tail cap).

I built this branch (`53dde36`) on Windows 11 and ran it as my daily driver in Obsidian: **19 real Antigravity turns**, 10 of them started while another turn was still in flight (two or three tabs at once), on a conversation history large enough that the old argv transport failed immediately with `ENAMETOOLONG`. The stdin transport holds up — no `ENAMETOOLONG`, no truncation, concurrent tabs do not interfere. One turn out of 19 failed, and it is the case in point 1 below.

Scope note so this is not read as more than it is: the longest of those 19 turns was 167 s, so the branch has **not** yet been exercised against the 5-minute boundary from #70 on my side. What I can confirm for #70 is that `--print-timeout` is passed and that the capability probe detects it on Windows (see point 2). The 10- and 11.6-minute turns I reported earlier were on my own local equivalent, not on this branch; I will report back once a genuinely long turn runs here.

Three field reports, since Windows is the part you could not cover.

### 1. `status:"ERROR"` can arrive together with a complete answer

This is the one thing I would change before merging. In the close handler:

```ts
if (streamResult?.status === 'ERROR') {
  reject(new Error(formatAntigravityResultError(streamResult)));
  return;
}
```

`agy` emits `status:"ERROR"` for a tool call its permission layer refuses, **after** the agent has already finished and streamed its answer. Two captures from today, both with `init.cwd` set to the vault and `--add-dir` in place:

```
result.status   = ERROR
result.response = 2716 chars      (write_to_file into a workspace subdirectory)
result.error    = declaring permissions: cortex tool write_to_file: convert tool call for
                  permissions: model output error: invalid tool call error (invalid_args)
                  D:\...\output\_tmp\build_batch3_clusters.py is not a valid artifact path;
                  artifacts must be in C:\Users\<user>\.gemini\antigravity-cli\brain\<id>/
```

```
result.status   = ERROR
result.response = 5073 chars      (view_file on a path that does not exist)
result.error    = declaring permissions: cortex tool view_file: ... failed to read file:
                  open D:/.../threads/455325.md: The system cannot find the file specified.
tools completed = run_command x29, view_file x6
```

In both runs the transcript ends with a full planner response, and the streamed `text_delta` frames sum to exactly `result.response` — so the answer is real and complete, and the current code throws it away in favour of a CLI-internal message the user can do nothing with.

This is not inferred from reading the diff: it reproduced on **this branch**, in Obsidian, on the run mentioned above. The user saw only

```
❌ Error: Antigravity CLI reported an error: declaring permissions: cortex tool write_to_file:
   ... D:\...\output\_tmp\populate_all_empty_clusters.py is not a valid artifact path ...
```

while `~/.gemini/antigravity-cli/brain/7118ef4b-…/logs/transcript.jsonl` ends at step 78 with a complete 1190-character `PLANNER_RESPONSE` summarising the finished work across 89 directories. The turn had done everything asked; only the report was withheld.

The timeout path already has the better behaviour ("A CLI that hangs right after emitting its result frame still delivered the answer"). I would mirror it here: when `status === 'ERROR'` but `response.trim()` is non-empty, resolve with the response — optionally with the error appended as a trailing note — and reject only when there is no answer. That is what I run locally now, and it turned a class of hard failures into a cosmetic warning.

Minimal repro on an empty temp directory, no vault involved:

```
agy --add-dir <tmp> --dangerously-skip-permissions \
    --input-format stream-json --output-format stream-json
stdin: {"event":"user","message":{"role":"user","content":"use write_to_file to create <tmp>\\sub\\x.py"}}
```

Result: `status:"ERROR"` with the artifact-path message, `response` says the file was created, and the file **is** on disk. Writing to a file directly in the workspace root succeeds instead, so a subdirectory target is one trigger; a `view_file` on a missing path is another. This looks like an `agy` defect rather than anything Grimoire can fix, which is exactly why the answer should survive it.

### 2. The capability probe is right to read stderr

Worth recording, because it is invisible on Linux and macOS: on the Windows build, `agy --help` writes **nothing to stdout**. Measured on this machine — stdout 0 bytes, stderr 2588 bytes, and `--add-dir` / `--input-format` / `--print-timeout` all appear only on stderr. `probeAntigravityCliCapabilities` collects both streams, so it detects everything correctly; I mention it only so the `child.stderr?.on('data', collect)` line does not get "simplified" later. A probe reading stdout alone would silently downgrade every Windows user to the legacy `--print` transport, which is the original #69 failure.

### 3. The provider unit tests do not pass on Windows

Not caused by this PR, but the PR widens the gap:

| Branch | Antigravity unit tests |
|---|---|
| `main` (1.1.8) | 7 failed, 67 passed |
| `fix/antigravity-stream-stdin-timeouts` | 16 failed, 96 passed |

Every failure is the same shape — the fixtures assume a POSIX launch:

```
- "/bin/zsh",
- ["-lc", "exec \"$0\" \"$@\"", "/usr/local/bin/agy", "--help"]
+ "cmd.exe",
+ ["/d", "/s", "/c", "\"/usr/local/bin/agy\" \"--help\""]
```

`buildAntigravityProcessLaunch` is behaving correctly; the expectations are hardcoded to the non-Windows branch, and `getResolvedProviderCliPath` returns `/usr/local/bin/agy`, which on Windows is not `.exe` and therefore takes the `cmd.exe` path. `AntigravityCliResolver.test.ts` already handles this properly by deriving fixtures from `process.platform`, so the pattern to follow exists in-tree. Happy to send a separate PR that makes the runtime and capabilities fixtures platform-aware — it is mechanical, and it would give the Windows-specific provider real CI coverage on the platform it targets.

Environment for all of the above: Windows 11, Node 22.23.2, npm 12.0.2, Obsidian desktop, `agy` Windows build 2026-08-20, CLI path pointing at `agy.exe`.

sandsaber:
Thank you — this is exactly the field testing we could not do ourselves, and the quality of the report (wire captures, transcript cross-checks, a minimal repro) made each point trivial to act on.

**Point 1 — applied in d4e594b3.** A non-empty `result.response` now resolves the turn regardless of `status` or the exit code, and the CLI complaint is appended as a trailing `> Warning:` note (capped at 500 chars) rather than replacing the answer. Only an ERROR result *without* an answer is still a hard failure. The same rule is mirrored in the timer path, so a hung CLI that already answered resolves too — that also means your SUCCESS-frame-with-exit-1 capture now resolves instead of rejecting, since a complete streamed answer beats the exit code there as well. New tests pin both the close-path and the hang-path variants, and `print.resultErrorWithAnswer` is recorded in the debug log when the note is attached. We also kept log cleanup on those turns: an answer means the run succeeded from the user's perspective.

**Point 2 — recorded.** The Windows help-only-on-stderr measurement is now a comment directly above the two `on('data', collect)` lines in `AntigravityCliCapabilities` and a bullet in the provider `AGENTS.md`, so a future "simplification" has to argue with the documentation first.

**Point 3 — please do send that PR.** Making the runtime and capabilities fixtures platform-aware (following the `AntigravityCliResolver.test.ts` pattern) is the right split: it is mechanical, it is pre-existing debt rather than this branch's regression, and it gives the provider real CI coverage on the platform it exists for. A separate PR also keeps the review trail honest about what changed here vs. what was widened.

One ask for your next daily-driver round: a genuinely long turn past the 5-minute boundary on this branch would close the last open question from #70 — the timers and `--print-timeout 29m` are in place and probe-verified on your machine, but a real >5m turn is the confirmation we are still missing. Thanks again for running 19 turns and reporting back with evidence.

greenkms:
Second daily-driver round on `53dde36`: **61 more turns** overnight, mostly two or three tabs in parallel. Every one of the 61 ended with a complete `PLANNER_RESPONSE` in `transcript.jsonl`. Below: the >5m confirmation you asked for, four more `status:"ERROR"` strings, and one inactivity-timer case worth a second look.

### The >5-minute confirmation for #70

Four turns crossed the old wall:

| conversation | wall clock |
|---|---|
| `01cfe98c` | 376 s |
| `4fcf970a` | 327 s |
| `b717e8b4` | 314 s |
| `b6d1b9e9` | 312 s |

Under the old absolute `PRINT_TIMEOUT_MS` all four would have been killed at 300 s. Three of them completed and delivered their answers normally, so the timeout rework does what it was meant to do. The fourth is the case below.

### One inactivity kill that I would not call a hang

`b717e8b4` ran 314 s and produced a 1806-character answer, but the user saw

```
❌ Antigravity request timed out after 5 minutes without CLI output.
```

File timestamps inside `brain/b717e8b4-…/` show why: a single silent stretch of **299 s** between two writes, then the answer. The turn was one long tool call with no intermediate frames, so from Grimoire's side it looked identical to a hang — and it was killed roughly 15 seconds before it would have delivered.

That is uncomfortably close to the threshold, and the workload is not exotic (one `run_command` that greps a large corpus). Two options, either is fine by me:

- raise `PRINT_INACTIVITY_TIMEOUT_MS` to something like 10 minutes, keeping the 30-minute absolute ceiling as the real backstop; or
- treat writes to the `--log-file` as liveness in addition to stdout/stderr, since `agy` keeps logging during a long tool call even when it emits no NDJSON frames.

Worth noting for the reasoning: in stream-json mode `agy` emits frames on step transitions, not continuously, so "no CLI output for N minutes" genuinely does not imply "hung" during a single long step.

### Four more ERROR strings, all with a complete answer behind them

All four surfaced as `Antigravity CLI reported an error: …` and, on `53dde36`, all four discarded a finished answer:

| trigger | transcript ends with |
|---|---|
| `cortex tool write_to_file … is not a valid artifact path` | 1190 chars |
| `cortex tool view_file … failed to read file: … cannot find the file specified` | 5073 chars |
| `cannot kill task "…/task-61": task is not running (status: DONE)` | 1578 chars |
| `another active schedule task "…/task-126" has a conflicting early termination condition "task-123"` | 286 chars |
| `context canceled` | complete response present |

So this is not a narrow `write_to_file` quirk: refused tool arguments, a stale task kill, a scheduler conflict and a cancellation all take the same path — `agy` reports a run-level error after the agent has already answered. That is precisely the class d4e594b3 now neutralises, which makes the change look even better justified than when I raised it.

I have switched my install to `d4e594b3` and will report back on how the trailing `> Warning:` note reads in practice, plus whether anything still slips through as a hard failure.

Environment unchanged: Windows 11, Node 22.23.2, npm 12.0.2, Obsidian desktop, `agy` Windows build, CLI path `agy.exe`.

sandsaber:
Verified on a real Windows machine (Windows 11 ARM64, `agy` 1.1.18, default `cmd.exe` launch path):

- `agy --help` advertises `--add-dir`, `--input-format`/`--output-format`, and `--print-timeout`, so the capability probe selects the stream-json transport.
- Stream-json transport works end-to-end: a 60k-character prompt written as a single `{"event":"user", ...}` stdin line completes a full turn and resolves the `{"event":"result"}` frame with `status:"SUCCESS"` and the model text — no `ENAMETOOLONG` (#69).
- Ordering pitfall confirmed live: passing `--print` before `--input-format` makes `agy` consume `--input-format` as the prompt and exit with code 2 — this PR correctly omits `--print` in stream-json mode.
- Cancel path: SIGTERM to the `cmd.exe` wrapper during an active turn left **no orphaned `agy.exe` processes** (checked via `tasklist` 5s after close), so the deferred `taskkill /T` concern did not reproduce here.
- Unit note: the `antigravity` suites have some Windows-local failures, but they fail identically on `main` (POSIX-shell mocks like `/bin/zsh` plus a `storage.getItem` mock issue), so they are pre-existing and not a regression from this PR.

sandsaber:
Thanks for the second round — both suggested remedies for the `b717e8b4` silent-step case are implemented in #73: the inactivity threshold is raised to 10 minutes (30-minute absolute ceiling and `--print-timeout 29m` unchanged), and growth of the `--log-file` now counts as liveness, polled every 15 s, with each observed growth logged as a `print.logLiveness` debug event so a silent-but-alive run is distinguishable from a hang in diagnostics. The inactivity error message now reads "without CLI output or log activity".

The four new ERROR-strings-with-complete-answers are exactly the class d4e594b neutralises (the streamed answer wins, the CLI complaint becomes a trailing `> Warning:` note) — good to have them confirmed across refused tool arguments, a stale task kill, a scheduler conflict, and a cancellation. Looking forward to your read on how the warning note behaves in practice.

PR #72, opened 2026-08-22
fix(opencode): launch acp with config content only

sandsaber:
## Summary

On Windows, `opencode acp` never completes `initialize` (and sometimes crashes with `0xC0000005`) whenever a file-based `OPENCODE_CONFIG` is present — even a trivial `{}` file, with either path separator style. Verified live against opencode 1.18.21 on Windows 11 ARM64. Since both the chat runtime and the aux query runners passed `OPENCODE_CONFIG`, the OpenCode provider was fully broken on Windows: no model list, no turns.

- The managed config now travels exclusively as `OPENCODE_CONFIG_CONTENT`. It is already complete: the user's `OPENCODE_CONFIG` base (if configured in Grimoire env settings) is read and merged during artifact preparation, so nothing is lost.
- `OPENCODE_CONFIG` is explicitly deleted from the child env so a user-configured value (from Grimoire environment settings or the parent process) cannot leak through and reintroduce the hang.
- `{file:...}` prompt references resolve relative to a config file that no longer reaches the CLI, so the managed system prompt is now inlined into the agent `prompt` fields. `system.md` is still written next to the config for inspection and stays byte-identical to the inlined prompt.
- **Kimi Code** (`src/providers/kimicode/`, the same OpenCode-CLI family) received the identical `KIMICODE_CONFIG` → `KIMICODE_CONFIG_CONTENT` treatment — it kept the exact Windows-hang launch shape this PR removes.
- MiMoCode mirrors the OpenCode CLI launch contract and gets the same `MIMOCODE_CONFIG` → `MIMOCODE_CONFIG_CONTENT` treatment per the provider-mirroring rule.

## Live verification (Windows 11 ARM64, opencode 1.18.21)

- `OPENCODE_CONFIG` set (any content, any path style) → `initialize` never answers; sometimes exit `0xC0000005`.
- `OPENCODE_CONFIG_CONTENT` only → `initialize` + `session/new` work; the `model` config option returns 12 models; a custom managed agent registers; a turn completes.
- With the session mode switched to the managed agent (`session/set_config_option`), a marker word embedded in the inlined system prompt is repeated by the model — the inlined prompt is genuinely loaded.

## Multi-review round

Two independent review agents (correctness/regressions + tests/contract) found no blockers; their findings are addressed in 20f894b:

- Aux query runners had zero child-env coverage — added env-contract tests for all three providers (a reverted `delete` now fails the suite at every launch site).
- Prepare tests parsed the on-disk config file; they now assert on `result.configContent` (the string the CLI actually consumes) and cross-check the inlined prompt against `system.md`.
- Kimi Code carried the same broken launch shape — fixed in this PR (see above).
- Known notes, deliberately not code: the 4-agent chat config inlines the prompt four times (~17.5 KB baseline, ~54% of the conservative 32,767-char Windows env-var budget; a custom system prompt over ~4 KB per agent could exceed it, though CreateProcessW/libuv measured delivering 200 KB fine); a user-set `OPENCODE_CONFIG_CONTENT`/`MIMOCODE_CONFIG_CONTENT`/`KIMICODE_CONFIG_CONTENT` in Grimoire env settings is intentionally overridden by the managed value (which already contains the user's base config).

## Testing

- Launch tests per provider (chat runtime + aux runner): child env carries `*_CONFIG_CONTENT` and never the file-based `*_CONFIG`, even when user env vars set it.
- Managed-config tests updated from `{file:...}` expectations to the inlined prompt text, asserted byte-identical to `system.md` via `configContent`.
- Local gate on Windows: unit suites for opencode/mimocode/kimicode — 49 failing vs 50 on `main` (all pre-existing Windows-only mock failures) with 5 new tests passing; full unit, integration, `typecheck`, `lint`, `build:release` including Obsidian community-review gates all pass. (MiMoCode and Kimi Code CLIs were not available locally for live checks; both changes are symmetric mirrors covered by mirrored unit tests.)

### DIFF
--- a/src/providers/kimicode/runtime/KimicodeAuxQueryRunner.ts
+++ b/src/providers/kimicode/runtime/KimicodeAuxQueryRunner.ts
@@ -203,7 +203,6 @@ export class KimicodeAuxQueryRunner implements AuxQueryRunner {
     this.reset();
     await this.startProcess({
       command: resolvedCliPath,
-      configPath: artifacts.configPath,
       configContent: artifacts.configContent,
       cwd,
       runtimeEnv,
@@ -241,7 +240,6 @@ export class KimicodeAuxQueryRunner implements AuxQueryRunner {
 
   private async startProcess(params: {
     command: string;
-    configPath: string;
     configContent: string;
     cwd: string;
     runtimeEnv: NodeJS.ProcessEnv;
@@ -249,10 +247,14 @@ export class KimicodeAuxQueryRunner implements AuxQueryRunner {
     const processEnv: NodeJS.ProcessEnv = {
       ...process.env,
       ...params.runtimeEnv,
-      KIMICODE_CONFIG: params.configPath,
       KIMICODE_CONFIG_CONTENT: params.configContent,
       PATH: params.runtimeEnv.PATH,
     };
+    // A file-based KIMICODE_CONFIG makes the acp mode hang or crash outright
+    // on Windows (Kimi Code mirrors the OpenCode CLI); the merged config must
+    // only travel as KIMICODE_CONFIG_CONTENT. Delete it explicitly so a
+    // user-configured value cannot leak through.
+    delete processEnv.KIMICODE_CONFIG;
 
     this.process = new AcpSubprocess({
       args: ['acp'],
--- a/src/providers/kimicode/runtime/KimicodeChatRuntime.ts
+++ b/src/providers/kimicode/runtime/KimicodeChatRuntime.ts
@@ -349,7 +349,7 @@ export class KimicodeChatRuntime implements ChatRuntime {
       }
       await this.startProcess({
         command: resolvedCliPath,
-        configPath: artifacts.configPath,
+        configContent: artifacts.configContent,
         cwd,
         runtimeEnv,
       });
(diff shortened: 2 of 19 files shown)


sandsaber:
Multi-review round complete (two independent reviewers: correctness/regressions and tests/contract). No blockers. Findings addressed in 20f894b:

1. **Aux-runner env contract untested** → env-contract tests added for all three providers' auxiliary runners; empirically, reverting the `delete processEnv.*_CONFIG` line now fails the suite at every launch site.
2. **Kimi Code kept the broken launch shape** (`KIMICODE_CONFIG` + `{file:...}` prompts, same OpenCode-CLI family) → same content-only fix applied, plus a new `KimicodeAcpLaunch` test mirroring opencode/mimocode.
3. **Prepare tests asserted on the on-disk config file** → now parse `result.configContent` (what the CLI consumes) and cross-check the inlined prompt against `system.md`.

Documented non-code notes: the 4-agent chat config rides at ~54% of the conservative 32,767-char Windows env-var budget (unbounded custom system prompt is the only way past it; 200 KB measured fine through CreateProcessW/libuv), and a user-set `*_CONFIG_CONTENT` in Grimoire env settings is intentionally overridden by the managed merged config.

PR #74, opened 2026-08-22
test(antigravity): make runtime and capabilities fixtures independent of the launch platform

greenkms:
## Problem

On Windows, the `antigravity` provider's unit tests fail broadly — not
from a behavior regression, but because several fixtures hardcode a
POSIX process-launch shape. Flagged during review of #71 (7 failed on
`main`/1.1.8 vs 16 on that PR's branch) and requested explicitly:

> **Point 3 — please do send that PR.** Making the runtime and
> capabilities fixtures platform-aware (following the
> `AntigravityCliResolver.test.ts` pattern) is the right split: it is
> mechanical, it is pre-existing debt rather than this branch's
> regression, and it gives the provider real CI coverage on the
> platform it exists for.
> — https://github.com/sandsaber/Grimoire/pull/71#issuecomment-5370169369

No issue number was opened for this; it's tracked only in that review
thread.

## Root Cause

`buildAntigravityProcessLaunch` wraps non-`.exe` commands on Windows in
`cmd.exe /d /s /c "<quoted line>"` instead of spawning `agy` directly.
The fixtures were written for the POSIX login-shell shape only:

- `AntigravityChatRuntime.test.ts` matched the print/probe child
  processes via `args.includes('--help')`, which is `false` once the
  flag is nested inside one quoted cmd.exe argument string — the probe
  then received the wrong child, never resolved, and every downstream
  stream-json assertion failed for a reason unrelated to the behavior
  under test.
- `AntigravityCliCapabilities.test.ts` asserted the exact spawn call
  (`'/bin/zsh', ['-lc', ...]`) instead of deriving it from the
  platform, so the one launch-shape assertion failed outright.

`AntigravityCliResolver.test.ts` already handles this correctly by
branching fixtures on `process.platform`.

## Approach

Rather than branching each assertion on `process.platform` (the
`AntigravityCliResolver.test.ts` pattern), this normalizes both
directions:

- `AntigravityChatRuntime.test.ts`: added `toAgyArgs()` /
  `parseWindowsShellCommandLine()`, which undo whichever launch
  wrapping was actually applied (POSIX login shell, Windows cmd.exe, or
  direct exec) and recover the logical `agy` arguments. Existing
  `args.includes('--help')` call sites, the `--log-file` path lookups,
  and the print-args filter now all go through this.
- `AntigravityCliCapabilities.test.ts`: the one launch-shape assertion
  now computes its expectation by calling `buildAntigravityProcessLaunch`
  directly instead of hardcoding a POSIX shape, so it can't drift from
  what the launch builder actually decides.

I chose normalization over per-assertion platform branching because
most of the affected assertions only care about the logical `agy`
flags, not the wrapping — normalizing once keeps the test bodies free
of `if (process.platform === 'win32')` noise. Happy to switch to the
branching style instead if you'd prefer consistency with
`AntigravityCliResolver.test.ts`.

Test-only change; no production code touched.

## Architecture

- [x] Provider-native behavior remains inside `src/providers/antigravity/`
      — N/A, no `src/` change at all, tests directory only.
- [x] N/A — no sibling providers affected (Antigravity-only fixtures).

## Security And Privacy

No security boundary change — test-only, no production code paths
altered.

## Verification

Automated commands run (Windows 11, Node 22.22.2, npm 12.0.2 — matching
CI's pinned toolchain, run through an isolated portable install so the
system Node/npm were not touched):

```text
npm run typecheck
npm run lint
npm run review:source
npm run test -- --selectProjects unit
npm run build:release
```

On Windows, `tests/unit/providers/antigravity` goes from 18 failing to
3. The remaining 3 are pre-existing and unrelated to process launch
(two involve a log-file unlink timing difference, one is
model-discovery ordering — separate root causes, out of scope here).
On the full unit suite: 83 → 70 failures (13 repaired), and no suite
enters the failing set — verified against a `main` baseline run with
the same toolchain immediately before and after.

Manual verification: Not run (test-only change; no runtime behavior to
exercise manually).

## Generated Artifacts

- [x] `npm run build:release` recreates `dist/grimoire/main.js`,
      `manifest.json`, and `styles.css` from the submitted source
      (verified locally; not included in this diff).
- [x] Generated root bundles, `dist/grimoire`, and local-only artifacts
      are not included.

## Review Checklist

- [x] The PR is focused on one coherent problem (platform-dependent
      test fixtures).
- [x] N/A — test-only change, no behavior change requiring new
      regression tests.
- [ ] N/A — no protocol payloads or error codes involved; this is a
      process-launch fixture, not provider protocol handling.
- [x] The description distinguishes automated tests (above) from
      manual verification (not run).

### DIFF
--- a/tests/unit/providers/antigravity/AntigravityChatRuntime.test.ts
+++ b/tests/unit/providers/antigravity/AntigravityChatRuntime.test.ts
@@ -114,10 +114,70 @@ function writeStreamJsonResult(
   writeStreamJsonFrame(proc, { event: 'result', result });
 }
 
+/**
+ * Undo the quoting `buildWindowsShellCommandLine` applies, so assertions can
+ * read the agy flags out of a `cmd.exe /d /s /c "<line>"` launch the same way
+ * they read them out of a POSIX one. Doubled trailing backslashes are left as
+ * written; no fixture path ends in one.
+ */
+function parseWindowsShellCommandLine(line: string): string[] {
+  const parsed: string[] = [];
+  let current = '';
+  let quoted = false;
+  let started = false;
+  for (let index = 0; index < line.length; index += 1) {
+    const char = line[index];
+    if (char === '"') {
+      if (quoted && line[index + 1] === '"') {
+        current += '"';
+        index += 1;
+      } else {
+        quoted = !quoted;
+        started = true;
+      }
+      continue;
+    }
+    if (char === ' ' && !quoted) {
+      if (started) {
+        parsed.push(current);
+        current = '';
+        started = false;
+      }
+      continue;
+    }
+    current += char;
+    started = true;
+  }
+  if (started) {
+    parsed.push(current);
+  }
+  return parsed;
+}
+
+/**
+ * Normalizes every launch shape `buildAntigravityProcessLaunch` can produce
+ * (POSIX login shell, Windows cmd.exe wrapper, direct exec) back to the agy
+ * arguments themselves, so the fixtures assert behavior instead of the host
+ * platform's process-launch convention.
+ */
+function toAgyArgs(spawnArgs: string[]): string[] {
+  if (spawnArgs[0] === '-lc' && spawnArgs[1] === 'exec "$0" "$@"') {
+    return spawnArgs.slice(3);
+  }
+  if (spawnArgs[0] === '/d' && spawnArgs[1] === '/s' && spawnArgs[2] === '/c') {
+    return parseWindowsShellCommandLine(spawnArgs[3] ?? '').slice(1);
+  }
+  return spawnArgs;
+}
+
+function isHelpProbeArgs(spawnArgs: string[]): boolean {
+  return toAgyArgs(spawnArgs).includes('--help');
+}
+
 function getSpawnedAgyArgs(): string[] {
   const printCall = mockedSpawn.mock.calls
     .map(([, args]: [string, string[]]) => {
-      const flagArgs = args[0] === '-lc' && args[1] === 'exec "$0" "$@"' ? args.slice(3) : args;
+      const flagArgs = toAgyArgs(args);
       const isPrint = flagArgs.includes('--print') || flagArgs.includes('--input-format');
       return { args: flagArgs, isPrint };
     })
@@ -127,7 +187,8 @@ function getSpawnedAgyArgs(): string[] {
 
 function getPrintLogFilePath(): string {
   const [, args] = mockedSpawn.mock.calls[mockedSpawn.mock.calls.length - 1] as [string, string[]];
-  return args[args.indexOf('--log-file') + 1];
+  const agyArgs = toAgyArgs(args);
+  return agyArgs[agyArgs.indexOf('--log-file') + 1];
 }
 
 // FIFOs exist on POSIX only; the drain-race test parks transcript recovery
@@ -491,7 +552,7 @@ describe('AntigravityChatRuntime', () => {
         prompt: 'Hello from transcript',
         runtimeEnv: process.env,
       });
-      const spawnArgs = mockedSpawn.mock.calls[0][1] as string[];
+      const spawnArgs = toAgyArgs(mockedSpawn.mock.calls[0][1] as string[]);
       const logFileArgIndex = spawnArgs.indexOf('--log-file');
       expect(logFileArgIndex).toBeGreaterThanOrEqual(0);
       const logFilePath = spawnArgs[logFileArgIndex + 1];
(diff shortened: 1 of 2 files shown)


sandsaber:
@greenkms, thank you for assisting in resolving this issue! Appreciated!

sandsaber:
Reviewed and merged-in-spirit — the approach is right, and to answer your open question directly: **normalization over per-assertion branching was the correct call here.** Fourteen call sites wrapped in `if (process.platform === 'win32')` would have buried the actual assertions. `AntigravityCliResolver.test.ts` branches because it has three platform-dependent assertions; this file has one platform-dependent *concept* and many assertions that don't care about it. Different shape, different tool.

I've pushed one commit on top (`586059a`) rather than round-tripping a review comment. What and why:

### The parser wasn't inverting `quoteWindowsArgument` completely

`parseWindowsShellCommandLine` handled the `""` quote escape but never undid the backslash doubling from:

```js
const prepared = value
  .replace(/(\\*)"/g, '$1$1""')
  .replace(/(\\+)$/, '$1$1');
```

Your comment called this out honestly ("no fixture path ends in one"), and that's true today. What made me want to close it anyway is that the affected arguments aren't fixed strings — `--print <prompt>` and `--add-dir <vaultPath>` put caller-controlled text straight into the argument list:

- `AntigravityChatRuntime.ts:899` — `args.push('--print', spec.prompt)`
- `AntigravityChatRuntime.ts:875` — `args.push('--add-dir', spec.addDirPath)`

The current vault fixture is `/tmp/grimoire-antigravity-test-vault`, POSIX-shaped even on Windows, so nothing bites. But the first fixture to use a realistic Windows vault path ending in a separator would fail on Windows only, with a diff pointing at the wrong thing — which is exactly the failure mode this PR set out to remove.

The fix consumes a backslash run as a unit and halves it when a quote follows, mirroring the producer's rule:

```ts
if (char === '\\') {
  let backslashes = 0;
  while (line[index] === '\\') { backslashes += 1; index += 1; }
  // quoteWindowsArgument doubles the run that precedes a quote or the
  // closing quote; every other run reaches the child verbatim.
  const wasDoubled = line[index] === '"';
  current += '\\'.repeat(wasDoubled ? Math.floor(backslashes / 2) : backslashes);
  started = true;
  continue;
}
```

### The helpers are now pinned against the real producer

`toAgyArgs` and `parseWindowsShellCommandLine` had become ~40 lines of quoting logic with no test of their own — if they drift from `quoteWindowsArgument`, every fixture that depends on them fails confusingly on one platform. The new `launch-shape normalization helpers` block (32 cases) round-trips them against `buildWindowsShellCommandLine` and `buildAntigravityProcessLaunch` across all three launch modes, asserting `launchMode` alongside the recovered args so a case can't silently stop exercising the shape it was written for.

Worth noting for future work on this suite: **the Windows shapes are verifiable from any host.** `buildAntigravityProcessLaunch` takes `platform` as its fourth parameter and `buildWindowsShellCommandLine` is pure string logic, so `'win32'` behavior can be pinned on macOS or Linux. That's how I checked your original helpers before touching them — they were correct on 29 of 33 cases, and the 4 failures were precisely the two documented lossy shapes counted twice.

### Verification (macOS, on top of your branch)

| | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm run lint` | clean |
| `tests/unit/providers/antigravity` | 137 passed (+32), failure set byte-identical to a `main` baseline |
| full unit suite | 7102 passed, same 10 failures as before the commit |

The 10 are pre-existing `setImmediate is not defined` failures in my local environment, present identically on `main` — unrelated to this branch either way.

### Left alone deliberately

**The CI matrix.** Your linked review framed this as giving the provider "real CI coverage on the platform it exists for," and that part isn't delivered yet: `.github/workflows/ci.yml` is `ubuntu-latest` only. Since your own numbers put the full Windows unit suite at ~70 remaining failures, adding `windows-latest` now would just paint CI red. I'll open a separate issue to work those down and then add the matrix entry — without it these fixtures will drift back, so it's worth tracking rather than assuming.

**`expect.objectContaining({ shell: launch.shell })`** in the capabilities test. `launch.shell` is `false` in every branch of `buildAntigravityProcessLaunch`, so it's a no-op versus the previous hardcoded `false` — but reading it off the launch object states the intent better, so it stays.

One thing the capabilities change does trade away: deriving the expectation from the same function production calls makes that assertion self-referential, so it would no longer catch a regression *inside* `buildAntigravityProcessLaunch`. That's fine here only because `AntigravityProcessLaunch.test.ts` already pins the concrete shapes (POSIX login shell, win32 direct `.exe`, cmd.exe quoting) independently. Flagging it so the reasoning is on the record rather than rediscovered later.

Thanks for splitting this out of #71 and for the platform-baseline numbers in the description — the before/after framing made the review much faster.

PR #76, opened 2026-08-22
test: derive path expectations from the platform instead of POSIX literals

sandsaber:
## Problem

The `windows-unit` job added in #75 reported 77 failing tests on `main`. Merging #74 removed 15 of them, leaving 62.

This PR takes the first and largest class: fixtures that assert a POSIX literal against a value the production code deliberately composes per platform. Every one of them could only ever pass on POSIX, and none of them indicate a defect in the code under test.

No issue number — the triage list lives in #75's description and in the `windows-unit-log` artifact.

## Root Cause

Three distinct mistakes, all with the same shape:

1. **Literal expectations for computed paths.** `normalizePathForFilesystem` routes through `path.win32.normalize` on Windows by design, so `/usr/local/bin` correctly returns `\usr\local\bin`. `resolveWorkspacePath` runs every input through `path.resolve`, so on Windows the workspace root carries the current drive. `encodeVaultPathForSDK` resolves before encoding — which mirrors how Claude Code names its own project directories — so the drive ends up in the encoded name.

2. **A fixture that measured something other than its subject.** `parsePathEntries`' split test used single-letter segments, and `translateMsysPath` reads a single-letter root as an MSYS drive spec, so `/a` arrived as `A:`. The translation is correct; the test claimed to be about separator splitting and was silently asserting MSYS behavior instead.

3. **A mock that could not match.** The SDK sidecar hydration test routed its own `readFile` mock with `p.includes('subagents/agent-ae5eb9a.jsonl')`. That never matches a `\`-separated path, so the mock fell through to the main transcript branch and the subagent's first tool call came back as `Task` instead of `Grep`. This one was not a formatting mismatch — the assertion was reading genuinely wrong data.

## Approach

Expectations are now derived the same way the production code derives its values — `path.join`, `path.resolve`, `path.normalize` — rather than spelled out for one platform. Where a fixture needs an absolute root, it starts from a platform-native one so `path.resolve` is a no-op; a POSIX literal would otherwise pick up whichever drive the runner happened to be on, which is not a stable fixture.

For `encodeVaultPathForSDK` that leaves the drive prefix as the only platform-varying part of the expected name, so the tests still assert concrete strings (`-Users-test-vault` / `C--Users-test-vault`) rather than restating the implementation.

The one case where the input itself was the problem — `parsePathEntries` — uses multi-character segments and carries a comment explaining why, so it does not get "simplified" back.

**No production code is changed.** In every case the behavior under test was already correct.

Verified on macOS that these values match what Windows actually produces: `path.win32.normalize` reproduces the exact strings the CI log recorded, and the encoded-name parity (`C--` + the POSIX tail) was checked against `path.win32.resolve` before the expectations were written.

## Architecture

- [x] N/A — tests only, no `src/` change.
- [x] N/A — no provider behavior changed; the Claude-owned fixtures stay under the Claude provider's test tree.

## Security And Privacy

- [x] N/A — no provider, session, or model data handled.
- [x] N/A — no permission or authorization behavior changes.
- [x] Filesystem, process, credential, MCP, and external-path boundaries are unchanged.
- [x] Fixtures contain no secrets or real vault contents.

Security notes: no security boundary change. Worth noting that `resolveWorkspacePath`'s containment tests were **not** weakened — the escape cases (`../../etc/hosts`, absolute paths outside the workspace, the symlink escape) were already passing on Windows and are untouched. Only the success-path expectations changed.

## Verification

Automated commands run (macOS):

```text
npm run test -- --selectProjects unit --testPathPatterns "utils/path|acp/resolveWorkspacePath"
npm run test -- --selectProjects unit --testPathPatterns "utils/sdkSession"
npm run typecheck
npm run lint
```

224 tests pass across the three touched suites; typecheck and lint are clean.

Manual verification: Not run — test-only change with no runtime behavior to exercise.

The real check is the `windows-unit` job on this PR. Expected: **62 → 42**, with `normalizePathForFilesystem` (4), `parsePathEntries` (1), `resolveWorkspacePath` (5), and `sdkSession` (10) leaving the failure list, and no suite entering it.

## Generated Artifacts

- [x] N/A — no source change, so no bundle to regenerate.

## Review Checklist

- [x] The PR is focused on one coherent problem (POSIX-only path expectations).
- [x] N/A — no behavior change; these fixtures are themselves the regression tests.
- [x] N/A — no protocol payloads or error codes involved.
- [x] The description separates what was verified locally from what only the Windows job can confirm.

## Not in this PR

The remaining ~42 failures are deliberately left for follow-ups, because they are not the same kind of change:

- **Suspected real Windows bugs** — `GrokUsageMetadataStore` returning `null` for session cost, `AgentManager` classifying a vault agent as built-in, `permission request CSS` reading an empty string. These need `src/` changes and, for the provider ones, sibling-provider review.
- **Tests that should be skipped off POSIX** rather than made portable, such as `findClaudeCLIPath › on Unix/macOS`.
- **Remaining launch-shape and provider-runtime fixtures** across Grok, Qwen, Gemini, OpenCode, MiMoCode, and Kimi Code.

### DIFF
--- a/tests/unit/providers/acp/resolveWorkspacePath.test.ts
+++ b/tests/unit/providers/acp/resolveWorkspacePath.test.ts
@@ -4,16 +4,21 @@ import * as path from 'node:path';
 
 import { resolveWorkspacePath } from '@/providers/acp/resolveWorkspacePath';
 
-const CWD = '/tmp/grimoire-test-vault';
+// Resolved rather than written as a literal: resolveWorkspacePath runs every
+// input through path.resolve, so on Windows the workspace root carries the
+// current drive and separators the POSIX literal never would.
+const CWD = path.resolve('/tmp/grimoire-test-vault');
 
 describe('resolveWorkspacePath', () => {
   describe('when containment is enforced (safe/plan mode)', () => {
     it('resolves a relative path inside the workspace', () => {
-      expect(resolveWorkspacePath(CWD, 'notes/today.md')).toBe(`${CWD}/notes/today.md`);
+      expect(resolveWorkspacePath(CWD, 'notes/today.md')).toBe(path.join(CWD, 'notes', 'today.md'));
     });
 
     it('allows an absolute path that stays inside the workspace', () => {
-      expect(resolveWorkspacePath(CWD, `${CWD}/notes/today.md`)).toBe(`${CWD}/notes/today.md`);
+      const inside = path.join(CWD, 'notes', 'today.md');
+
+      expect(resolveWorkspacePath(CWD, inside)).toBe(inside);
     });
 
     it('allows the workspace root itself', () => {
@@ -66,12 +71,13 @@ describe('resolveWorkspacePath', () => {
 
   describe('when containment is disabled (active / full_access mode)', () => {
     it('returns an absolute path outside the workspace unchanged', () => {
-      expect(resolveWorkspacePath(CWD, '/etc/hosts', { allowOutsideWorkspace: true })).toBe('/etc/hosts');
+      expect(resolveWorkspacePath(CWD, '/etc/hosts', { allowOutsideWorkspace: true }))
+        .toBe(path.resolve('/etc/hosts'));
     });
 
     it('returns an escaping relative path resolved against the workspace', () => {
       expect(resolveWorkspacePath(CWD, '../sibling/file.md', { allowOutsideWorkspace: true })).toBe(
-        '/tmp/sibling/file.md',
+        path.join(path.dirname(CWD), 'sibling', 'file.md'),
       );
     });
   });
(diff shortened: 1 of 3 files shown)


sandsaber:
`windows-unit` on this PR: **62 → 42 failing tests**, 25 → 23 suites. Exactly the predicted number.

All four targeted clusters left the failure list — `normalizePathForFilesystem`, `parsePathEntries`, `resolveWorkspacePath`, `sdkSession` now report zero — and no suite entered it. The Linux `validate` job stayed green, so nothing regressed on the platform that was already covered.

Worth recording what this says about the workflow, since it was the open question when #75 was written: **none of this needed a Windows machine.** The expectations were derived on macOS from `path.win32`, checked against the exact strings the previous run's log recorded, and the job confirmed them 1m52s later. The prediction in the description was written before the job ran and did not need adjusting.

PR #78, opened 2026-08-23
test(utils): anchor CLI-discovery fixtures to the host platform

sandsaber:
## Problem

Second slice of the Windows failures the `windows-unit` job reports. #76 took the suites down to 42; this one takes the eleven remaining failures in the shared utility suites — `utils/path`, `utils/utils`, and `utils/env`.

They split into two genuinely different problems that happened to look alike in the log.

## Root Cause

**Five fixtures pinned a layout the code deliberately does not use on Windows.**

`getNpmClaudeCodeEntrypointPaths` probes `%USERPROFILE%\AppData\Roaming\npm` on Windows, not `~/.npm-global/lib`, because a global npm install has no `lib/` level there. Two fixtures hard-coded the POSIX parent, so on Windows they were proving that nothing exists at a path the resolver never consults — `Received: null` was the resolver behaving correctly.

The PATH fallback had the same shape from a different angle: it spliced `PATH` with a hard-coded `':'` and expected a bare `claude`, while `resolveClaudeFromPathEntries` looks for `claude.exe` before `claude` and `PATH` is delimiter-separated per platform.

Two `normalizePathForFilesystem` expectations were plain separator literals in a file whose neighbouring cases already build theirs with `path.normalize`.

**Six could not be repaired at all — only honestly scoped.**

`findClaudeCLIPath > on Unix/macOS` and `Obsidian CLI path integration` already override `process.platform` to `darwin`/`linux` and re-require the module. That is not enough, and the reason is worth stating plainly because it will come up again: **overriding `process.platform` does not redirect the `path` module.** `path.join` keeps emitting `\`-separated candidates and `path.delimiter` keeps returning `';'` on a Windows host, so POSIX resolution is unreachable there no matter what the platform value says. `getEnhancedPath` composes with both, so the macOS app-bundle and Linux AppImage expectations were comparing `\`-joined candidates against a `;`-delimited string.

## Approach

For the first group, expectations now derive the package parent, the binary name, and the delimiter from the host instead of spelling out one platform's answer. No production code changed — the resolver was right in every case.

For the second group, the tests are skipped on Windows through a `describeOnPosix` binding:

```ts
const describeOnPosix = process.platform === 'win32' ? describe.skip : describe;
```

Two notes on that choice. It is `describe.skip` rather than the `if (isWindows) return;` idiom already present elsewhere in these files, because an early return reports the test as **passed** — which is exactly the kind of quiet false signal the `windows-unit` job exists to remove. Skipped tests show as skipped. And the coverage is not lost: the Linux `validate` job runs all of them on every PR, which is the platform they describe.

Rewriting them to pass on Windows was considered and rejected. The only way to do it would be to normalize the expected macOS and Linux paths through the host's `path`, at which point the assertion is about neither platform.

## Architecture

- [x] N/A — tests only, no `src/` change.
- [x] N/A — no provider behavior changed.

## Security And Privacy

- [x] N/A — no provider, session, or model data handled.
- [x] N/A — no permission or authorization behavior changes.
- [x] Filesystem, process, credential, MCP, and external-path boundaries are unchanged.
- [x] Fixtures contain no secrets or real vault contents.

Security notes: no security boundary change. Worth flagging one thing a reviewer should push back on if they disagree: skipping the POSIX CLI-discovery tests on Windows means a Windows-only regression in that code path would not be caught by the Windows job. That is acceptable because the code path is explicitly `if (!isWindows)`-gated, and the Windows branch has its own coverage in the same file — but it is a deliberate reduction in where those assertions run, not a no-op.

## Verification

Automated commands run (macOS):

```text
npm run test -- --selectProjects unit --testPathPatterns "utils/env|utils/utils|utils/path"
npm run typecheck
npm run lint
```

276 tests pass across the three touched suites; typecheck and lint clean.

Manual verification: Not run — test-only change.

The `windows-unit` job on this PR is the real check. Expected: **42 → 31**, with five failures repaired and six moving to skipped. No suite should enter the failure list, and the Linux job should be unchanged since `describeOnPosix` resolves to plain `describe` there.

## Generated Artifacts

- [x] N/A — no source change.

## Review Checklist

- [x] The PR is focused on one coherent problem (host-platform assumptions in the shared utility fixtures).
- [x] N/A — no behavior change; these fixtures are the regression tests.
- [x] N/A — no protocol payloads or error codes involved.
- [x] The description separates what was verified locally from what only the Windows job can confirm.

### DIFF
--- a/tests/unit/utils/env.test.ts
+++ b/tests/unit/utils/env.test.ts
@@ -1216,7 +1216,14 @@ describe('getExtraBinaryPaths (Windows branches)', () => {
   });
 });
 
-describe('Obsidian CLI path integration', () => {
+// getEnhancedPath composes with path.join and path.delimiter, which stay bound
+// to the host no matter what process.platform is overridden to. On a Windows
+// runner these macOS- and Linux-specific expectations would be comparing
+// `\`-joined candidates against a `;`-delimited string, so they are skipped
+// rather than rewritten into an assertion about neither platform.
+const describeOnPosix = isWindows ? describe.skip : describe;
+
+describeOnPosix('Obsidian CLI path integration', () => {
   const originalPlatform = process.platform;
   const originalExecPath = process.execPath;
   const originalEnv = { ...process.env };
--- a/tests/unit/utils/path.test.ts
+++ b/tests/unit/utils/path.test.ts
@@ -413,6 +413,20 @@ describe('normalizePathForVault', () => {
   });
 });
 
+// getNpmClaudeCodeEntrypointPaths probes a different package parent per
+// platform — %USERPROFILE%\AppData\Roaming\npm on Windows, ~/.npm-global/lib
+// on POSIX — because global npm has no lib/ level on Windows. Anchoring a
+// fixture to the other platform's layout only proves that nothing is there.
+const NPM_PACKAGE_PARENT = isWindows
+  ? path.join(os.homedir(), 'AppData', 'Roaming', 'npm')
+  : path.join(os.homedir(), '.npm-global', 'lib');
+
+function claudeCodeEntrypoint(entrypoint: string): string {
+  return path.join(
+    NPM_PACKAGE_PARENT, 'node_modules', '@anthropic-ai', 'claude-code', entrypoint,
+  );
+}
+
 describe('findClaudeCLIPath', () => {
   afterEach(() => {
     jest.restoreAllMocks();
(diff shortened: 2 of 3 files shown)


sandsaber:
`windows-unit` on this PR: **42 → 31 failing, 8 skipped**. Exactly the predicted number. The Linux `validate` job is unchanged, as expected — `describeOnPosix` resolves to plain `describe` there.

Recording the one judgement call in this PR, since it is the only part a reviewer could reasonably want reversed.

The `on Unix/macOS` block holds six cases. Four failed on Windows; **two passed**, and skipping the block stops running those two there:

- `should return null when Claude CLI is not found`
- `should not return a directory path even if it exists`

They were skipped deliberately rather than left running. Both assert a null result, and on Windows the POSIX branch they exercise is unreachable — so they return null no matter what the resolver does. Keeping them would leave two green lights that cannot fail, which is the same false signal the job was added to remove. Linux still runs all six on every PR.

The reasoning is now in the file (`cf63340`), at the block and on the two cases, so it does not have to be reconstructed from this thread later.

The alternative — `it.skip` on the four failures only — is a few minutes of work if you would rather keep the two running on Windows.

PR #79, opened 2026-08-23
build: pin checkout line endings to LF so Windows builds reproduce the release bundle

greenkms:
## Problem

`npm run build:release` on Windows does not reproduce the published release bundle, and the difference is in shipped content rather than in formatting.

Building tag `1.1.9` (`c001881`) from a default Windows clone with the CI-pinned toolchain (Node 22.22.2, npm 12.0.2) yields:

| Bundle | SHA-256 |
|---|---|
| `dist/grimoire/main.js`, default Windows clone | `7383218053af0214866e0ff18fb968568c3601a75ebcc6dd369c86f9749d7eb5` |
| `main.js` published as the 1.1.9 release asset | `36e43950b4d4a9bb7204284b018f149036e85367f2382c1a5cf4ff856d2b747f` |

Same commit, same Node, same npm, different bytes — and the Windows bundle is 1 524 bytes larger.

This blocks the check `CONTRIBUTING.md` asks for ("Verify the release bundle from source before tagging") for anyone on Windows, and it makes "is this asset really built from this tag?" unanswerable for a Windows reviewer.

No issue number. Found while verifying my own build after #74.

## Root Cause

Two independent facts combine.

**1. A default Windows clone has a CRLF working tree.** Git for Windows ships `core.autocrlf=true` in its *system* config, so this is the out-of-the-box state for essentially every Windows contributor, not a personal setting:

```
$ git config --show-origin --show-scope --get-all core.autocrlf
system  file:C:/Program Files/Git/etc/gitconfig       true
$ git --version
git version 2.50.0.windows.1
```

The index is already clean — LF everywhere — but the checkout is not:

```
$ git ls-files --eol | awk '{print $1}' | sort | uniq -c    # index
   1110 i/lf
      4 i/-text
$ git ls-files --eol | awk '{print $2}' | sort | uniq -c    # working tree, fresh clone on Windows
   1110 w/crlf
      4 w/-text
```

**2. Exactly one working-tree file reaches the bundle as raw text.** Everything else in the repository passes through a parser that normalizes line terminators: the ECMAScript spec normalizes CRLF inside template literals (verified separately — minifying a CRLF source and its LF twin through this repo's own esbuild produced byte-identical output), and the CSS pipeline normalizes as well. The exception is `esbuild.config.mjs`:

```js
const changelogMarkdown = readFileSync('CHANGELOG.md', 'utf-8');   // line 33
…
  define: {
    GRIMOIRE_CHANGELOG_MARKDOWN: JSON.stringify(changelogMarkdown), // line 192
  },
```

`define` substitutes the literal at each surviving use site in `src/app/changelog/source.ts`, and esbuild emits the long multi-line string as a template literal. A raw LF inside a template literal costs one byte; a CR cannot stay raw there (the spec would normalize it away), so esbuild escapes each one as `\r` — two bytes, and a character that is now genuinely part of the shipped string.

The arithmetic closes exactly:

- `CHANGELOG.md` at 1.1.9 has 381 line endings;
- after minification two copies of the literal survive;
- escaped `\r` sequences in the bundle: **122** (LF build) → **884** (CRLF build) = **+762** = 2 × 381;
- bundle size delta: 762 × 2 bytes = **1 524** — the exact observed difference.

First differing byte (offset 3 476 731) is inside that literal:

```
LF build:    …function Hzt(){try{if(`# Changelog
             
             ## 1.1.9 - 2026-08-22
CRLF build:  …function Hzt(){try{if(`# Changelog\r
             \r
             ## 1.1.9 - 2026-08-22\r
```

So `getEmbeddedChangelogMarkdown()` returns markdown containing 381 CR characters when the plugin is built on Windows. I did not test whether the changelog renderer visibly misbehaves on that input — the point here is that the shipped string differs, not only the file hash.

## Approach

Add a `.gitattributes` that pins the working tree to LF for every tracked text file:

```
* text=auto eol=lf
```

`text=auto` leaves Git's own binary detection in charge, so the four PNGs under `assets/readme/` (already `i/-text`) are untouched. The repository contains no `.bat`, `.cmd`, `.ps1`, or `.sh` files that would want CRLF (`git ls-files` matches none), so no per-glob exceptions are needed.

**The change is content-neutral.** Every tracked text blob is already LF in the index, so adding the file rewrites nothing:

```
$ git add --renormalize .
$ git diff --cached --numstat | wc -l
0
```

**Alternative considered and rejected: normalize at the embed point** (`readFileSync('CHANGELOG.md', 'utf-8').replace(/\r\n/g, '\n')`). It is a smaller diff and it does fix this symptom. I did not propose it because it fixes one instance of a class: any future `readFileSync` of a working-tree file re-opens the same trap silently, and the Windows fixture work in #74/#76/#78 has been steadily uncovering neighbouring path- and separator-shaped assumptions. Pinning the checkout removes the input, not the one symptom. If you prefer the narrow fix, I am happy to swap it — the evidence above applies either way.

**Migration note for existing Windows clones.** `.gitattributes` governs future checkouts; a clone that already materialized CRLF keeps it until the tree is re-read once:

```
git rm --cached -r .
git reset --hard
```

Worth a line in `CONTRIBUTING.md` if you want it; I left the docs alone to keep this PR to one coherent change.

## Architecture

Architecture notes:

No source, provider, or shared contract is touched. This is repository-level checkout policy — the owning layer is the build/VCS configuration, and `.gitattributes` at the root is where Git looks for it.

## Security And Privacy

Security notes:

No security boundary change. The change is a checkout-time normalization of line endings for text files. It has one adjacent benefit worth naming: a reproducible bundle is what lets a reviewer confirm that a published release asset corresponds to the reviewed source, which today cannot be done from Windows.

## Verification

Automated commands run:

```text
# Fresh clone of 1.1.9 (c001881), default Windows settings, Node 22.22.2 / npm 12.0.2
npm ci
npm run build:release            # -> main.js sha256 73832180…  (does not match the release asset)

printf '* text=auto eol=lf\n' > .gitattributes
git add --renormalize .          # 0 files staged: content-neutral
git rm --cached -r . && git reset --hard
git ls-files --eol               # 1110 w/lf, 4 w/-text

npm run build:release            # -> main.js sha256 36e43950…  (byte-identical to the 1.1.9 release asset)
```

Both builds ran `prebuild:release` (`lint`, `review:source`, `review:css`, `review:deps`) and the bundle self-check ("Verified release bundle loads from isolated install dir" / "opens grimoire-view") without errors.

Reverting the experiment (removing `.gitattributes`, re-checking out to CRLF, rebuilding) reproduced `73832180…` again, so the result is not order-dependent.

Manual verification: Windows 11, Obsidian desktop. The 1.1.9 release asset currently installed in my vault hashes to `36e43950…`, which is what the LF build reproduces. No Obsidian workflow change is expected from this PR and none was exercised.

## Generated Artifacts

## Review Checklist

- [ ] Behavior changes have focused regression tests. — see the note below; no unit-testable surface.
- [ ] Protocol tests use real structured payloads and error codes. — not applicable, no protocol code is touched.

On the regression-test box: there is no unit-testable surface here — the defect lives in what Git materializes before any code runs, so the guard is the byte comparison above rather than a test file. If you would like it enforced, the natural place is the `windows-unit` job from #75: build the bundle on Windows and on Linux from the same commit and compare hashes. I did not add that here because it belongs to the CI conversation you are already running, not to this one-line policy change.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+# Git for Windows ships core.autocrlf=true in its system config, so a default
+# Windows clone materializes every tracked text file with CRLF while the index
+# stores LF. Multi-line template literals in src/ carry those bytes straight
+# into the minified bundle, so `npm run build:release` on Windows produces a
+# main.js that differs from the release CI builds on ubuntu-latest even with an
+# identical commit and toolchain. Pin the working tree to LF for everyone.
+* text=auto eol=lf


sandsaber:
@greenkms Thank you — this is an exceptional piece of work, and I want to be specific about why.

You didn't just report that the bundle differed, you closed the arithmetic. 381 line endings × 2 surviving copies of the literal × 2 bytes = 1 524, matching the observed size delta exactly, with the first differing byte located at offset 3 476 731 and quoted in context. That is a proof, not a hypothesis, and it turned what could easily have been several rounds of back-and-forth into a short review. The part most people would have skipped is the one that makes the argument airtight: actually verifying that the ES and CSS parsers normalize CRLF, so that `readFileSync('CHANGELOG.md')` in `esbuild.config.mjs` is demonstrably the *only* raw-text path into the bundle.

I also appreciate that you named the alternative you rejected (normalizing at the embed point), explained the reasoning — one instance of a class versus the class itself — and offered to swap if we preferred the narrow fix. And showing content-neutrality with `git add --renormalize .` *before* asking anyone to trust the change is exactly the evidence a reviewer needs to merge quickly.

Merging as is. Three small follow-ups I'll push myself so you don't have to spin another round:

- The in-file comment attributes the leak to multi-line template literals in `src/`, but your own analysis shows the parser normalizes those. Rewording it to point at the `CHANGELOG.md` → `define` path, so the next reader finds the real constraint instead of a decoy.
- Taking you up on the `CONTRIBUTING.md` line. As merged, a contributor already on a CRLF clone keeps it until the tree is re-read — which is precisely the group this fix exists for — so the migration steps you documented here belong in the repo.
- Adding a note that a fixture needing literal CRLF bytes must opt out with its own `-text` rule, since the Windows fixture work in #74/#76/#78 makes that a realistic near-term need.

Your read that this is a class of bug rather than a single instance is right, and it is the reason this is the correct fix rather than the smaller one. Thanks again for the care you put into this.

greenkms:
Confirmed on the comment — that line is stale rather than a disagreement. The
template-literal wording was written before the micro-experiment that disproved
it; when the measurement came back I rewrote the PR body and the commit message
and missed the file itself. Your rewording is the correct one: `readFileSync
('CHANGELOG.md')` → `define` is the only raw-text path into the bundle.

One measurement for the third follow-up, so you don't have to take it yourself.
On `c11b9bb`:

```
$ git ls-files --eol | awk '{print $1}' | sort | uniq -c
   1111 i/lf
      4 i/-text
```

Zero CRLF blobs in the index, so nothing in the tree needs the opt-out today.
The only suite that exercises CRLF handling — `tests/unit/utils/inlineEdit.test.ts`,
`handles \r\n line endings` — builds its input in source rather than reading a
file, so it is unaffected either way.

That matches your framing of the note as preventive, and it is worth writing
down for exactly that reason: the first fixture that does need literal CRLF
bytes would otherwise be normalized silently, with the assertion failing for a
reason that points somewhere else entirely.

PR #87, opened 2026-08-24
fix(claude): rebuild the plan panel from the sdk's task tools

sandsaber:
Corrects #86, which did not achieve what its description claimed.

## What #86 got wrong

#86 named `TodoWrite` in `allowedTools` to keep the plan panel working. Probing Claude Code 2.1.233 directly — starting sessions and reading the `tools` array off the SDK's `init` message — shows that was based on a misreading of the 0.3.233 notes.

**`TodoWrite` does not exist as a tool any more.** It was replaced by `TaskCreate` / `TaskGet` / `TaskUpdate` / `TaskList`. The release notes listed the old and new names together as one group; #86 read that as five tools to restore.

| Configuration | Tools | `TodoWrite` | `Task*` |
|---|---|---|---|
| `sonnet`, no options | 29 | no | no |
| `sonnet` + `allowedTools: ['TodoWrite']` (#86) | 33 | **no** | all four |
| `sonnet` + `CLAUDE_CODE_ENABLE_TODO_TOOLS=1` | 33 | no | all four |
| `sonnet` + `allowedTools` with all five names | 33 | no | all four |
| `claude-sonnet-4-5` (older model), no options | 33 | no | all four |
| stock (`settingSources: []`) | 27 | no | no |
| stock + `allowedTools: ['TodoWrite']` | 31 | no | all four |
| `tools: ['Read','Grep']` + `allowedTools: ['TodoWrite']` | 3 | no | no |

So #86 *did* re-enable the group — naming any member is enough, even the retired name — but the panel stayed empty, because the renderer keys on `TodoWrite` and nothing emits it any more.

The regression is also **wider than #86 assumed**: row 5 shows an older model has no `TodoWrite` either, so this was never limited to the newest models.

## Changes

**Name the real tools in the opt-in constant.** Functionally the same key, but the retired name read like a typo to anyone opening the file later.

**Fold the task calls back into the plan panel.** This is the actual fix. `TodoWrite` carried the whole list in one call. The task tools are incremental — a create carries no id until its result returns, and an update names only the fields it changes — so no single call ever describes the plan.

`claudeTaskPlanState` accumulates the calls and replays them as whole-list `TodoWrite` input. That is the same provider-neutral shape the Codex adapter already synthesizes from `update_plan` ([`CodexNotificationRouter.ts`](https://github.com/sandsaber/Grimoire/blob/main/src/providers/codex/runtime/CodexNotificationRouter.ts)), so nothing in `src/features/` changes and the existing renderer, panel, and conversation-restore paths are reused untouched.

Details worth a reviewer's attention:

- A `TaskList` result **replaces** the ledger rather than merging. It is the authoritative snapshot, so a task deleted elsewhere has to disappear from the panel.
- The replay reuses one tool-call id, so the plan stays a single updating entry instead of a new card per task call.
- Subagent plans are excluded — the panel tracks the main agent.
- Folding the same create twice is a no-op, because a streamed block surfaces as a tool use before the assembled message repeats it.

## The open question from #86 is now answered

The last table row settles it: **`allowedTools` does not re-add anything past a restricting `tools` list.** A slash command that restricts tools therefore gives up plan tracking. That is the command's own intent, so it is pinned by a test rather than worked around.

## Testing

Nine unit tests for the ledger, covering the deferred create, the `activeForm` fallback, status changes and ordering, deletion, an update for an unknown task, list-replaces-ledger, and double-folding. One asserts the replayed shape survives the shared `parseTodoInput` validator rather than merely looking similar. Two more at the transform level cover the emitted chunk and the no-state case.

`npm run typecheck`, `npm run lint` and `npm run build:release` are green, with no drift in generated artifacts.

Unit suite: 7124 passing, 9 failing. All 9 are `ReferenceError: setImmediate is not defined` in the two Antigravity suites, identical on a clean `main`, caused by running locally on Node 26 rather than CI's Node 22.

## Not covered by tests

The ledger is driven by the shapes in the SDK's `sdk-tools.d.ts` (`TaskCreateOutput`, `TaskListOutput`, and friends), not by a captured live transcript. The wiring is worth one manual run in the vault to confirm the panel actually populates.

### DIFF
--- a/src/core/tools/toolNames.ts
+++ b/src/core/tools/toolNames.ts
@@ -18,6 +18,13 @@ export const TOOL_SUBAGENT_LEGACY = 'Task' as const;
 // Kept as an alias while the internal codebase is still named around "Task".
 export const TOOL_TASK = TOOL_SUBAGENT;
 export const TOOL_TODO_WRITE = 'TodoWrite' as const;
+// Task tracking replaced TodoWrite in Claude Code 2.1.233. Grimoire still renders
+// plans through TOOL_TODO_WRITE, so provider adapters fold these into that shape.
+export const TOOL_TASK_CREATE = 'TaskCreate' as const;
+export const TOOL_TASK_GET = 'TaskGet' as const;
+export const TOOL_TASK_LIST = 'TaskList' as const;
+export const TOOL_TASK_UPDATE = 'TaskUpdate' as const;
+
 export const TOOL_TOOL_SEARCH = 'ToolSearch' as const;
 export const TOOL_WEB_FETCH = 'WebFetch' as const;
 export const TOOL_WEB_SEARCH = 'WebSearch' as const;
--- a/src/providers/claude/runtime/ClaudeChatRuntime.ts
+++ b/src/providers/claude/runtime/ClaudeChatRuntime.ts
@@ -75,6 +75,7 @@ import { encodeClaudeTurn } from '../prompt/ClaudeTurnEncoder';
 import { isContextWindowEvent, isSessionInitEvent, isStreamChunk } from '../sdk/typeGuards';
 import type { TransformEvent } from '../sdk/types';
 import { getClaudeProviderSettings } from '../settings';
+import { createClaudeTaskPlanState } from '../stream/claudeTaskPlanState';
 import {
   createTransformStreamState,
   createTransformUsageState,
@@ -270,6 +271,7 @@ export class ClaudeChatRuntime implements ChatRuntime {
   private turnMetadata: ChatTurnMetadata = {};
   private bufferedUsageChunk: StreamChunk & { type: 'usage' } | null = null;
   private streamTransformState = createTransformStreamState();
+  private taskPlanState = createClaudeTaskPlanState();
   private usageTransformState = createTransformUsageState();
 
   private getLegacyPluginDeps(): GrimoirePlugin & {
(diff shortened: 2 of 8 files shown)


sandsaber:
Verified end to end in the test vault. A forced run produced exactly the intended sequence:

```
[5] TaskCreate  {"subject":"проверка панели","description":"проверка панели"}
[6] TodoWrite   id: claude-task-plan
    {"todos":[{"content":"проверка панели","activeForm":"проверка панели","status":"completed"}]}
[7] TaskUpdate  {"taskId":"1","status":"completed"}
```

Four things this confirms beyond the unit tests:

- The task tools are reachable — the model located them through `ToolSearch` and called them.
- The ledger replays them as `TodoWrite`, so the panel is fed.
- The `activeForm` fallback fires: the model omitted the field and the subject was used.
- The stable id works as designed. There is a single `TodoWrite` entry, and it carries `status: completed` even though it appears *before* `TaskUpdate` in the recorded order — the entry was updated in place rather than duplicated per task call.

It also surfaced a defect the tests missed: that entry sat at `status: running`. The replay emitted the tool use but never the matching result, so the plan card keeps its spinner forever. The Codex adapter emits both halves; only the call had been carried over here.

Fixed in f16d116, with the result now pinned by the transform test — the previous assertions checked the chunk's contents but not that the call ever completes, which is why they stayed green.

PR #89, opened 2026-08-25
fix(claude): seed the catalog probe suppressor when the CLI path resolves late

greenkms:
## Symptom

A measured, reproducible cost on every plugin load. Owner's isolated
measurement: 5-hour plan window at 0% and not yet started (no requests at all)
→ launch Obsidian → wait for indexing → **2% consumed without opening a chat
tab, selecting a model, or sending anything**. Opening a Claude tab and sending
one message then took it to 7%.

PR #85 added a seed intended to prevent exactly this. It is installed and it
never worked.

## Root cause

`createClaudeModelCatalog` is called from `createClaudeWorkspaceServices`, which
runs *inside* `ProviderWorkspaceRegistry.initialize()`. The registry assigns the
services only after that promise resolves:

```ts
this.services[providerId] = await this.getWorkspaceRegistration(providerId).initialize({...});
```

So while the catalog is being constructed:

| call | during construction | at refresh |
| --- | --- | --- |
| `getServices('claude')` | `undefined` | services object |
| `getCliResolver('claude')` | `null` | resolver |
| `getResolvedProviderCliPath('claude')` | `null` → `?? ''` | real CLI path |

`cliPath` is part of `buildClaudeModelCatalogCacheKey`, so the seed was filed
under `cliPath: ""` while every subsequent `refreshModels` looked it up under the
real resolved path. The keys never matched, the seed was never found, and the SDK
probe ran on every plugin load — which is what the seed exists to prevent.

Confirmed in a real debug log (`.grimoire/logs/2026-08-25.jsonl`):

```
09:07:52.237Z plugin              loaded
09:08:04.041Z provider.claude     modelCatalog.refresh.empty     ← real SDK probe
09:08:54.796Z provider.claude     modelCatalog.refresh.skipped   ← reason: cache_fresh
```

`refresh.empty` is only written after an actual SDK call. The second call 50s
later hits the cache because by then both keys are refresh-time keys. Only the
seed key was the odd one out.

## Why it went unnoticed

Two independent blind spots:

1. **The existing reload test never modelled the real ordering.** Its mock
   returned `/claude` both at construction and at refresh, so the keys always
   matched and the seed always appeared to work.

2. **The plan-usage indicator cannot see this cost.** Every Claude
   `refresh.succeeded` in a full day of logs carries `pctKnown: false`; the SDK
   only sends `utilization` near the window ceiling (observed only in the 99–100%
   range). Earlier investigation concluded "the probe shows no pct movement,
   therefore it is free" — on an instrument that cannot distinguish 0% from 2%.

## Fix

When the CLI path cannot be resolved at construction, defer the seed to the first
`refreshModels` call, where the key is computed from exactly the same inputs the
lookup uses.

An eager seed is kept when the path *is* already resolvable, so a genuine
mid-session CLI path change still forces a re-probe — the existing
`still probes after a reload when the resolved CLI path changed` test passes
unchanged.

A new debug event `modelCatalog.refresh.skipped` with
`reason: 'seeded_on_first_use'` makes the seed observable, since the original
defect was invisible precisely because nothing distinguished "seeded" from
"probed".

## Tests

New: `suppresses the reload probe even when the CLI resolver is not reachable yet
at construction` — pins the production initialization order. It fails on `main`
(probe called once) and passes with this change.

Unit suite is at the repo's pre-existing Windows baseline: 22 failed suites / 35
failed tests before and after; passing 7069 → 7070 (+1 new test). `npm run
typecheck`, `npm run lint`, and `npm run build:release` all clean.

## Verified after the fix

The owner repeated the isolated protocol on a build of this branch (auto-ping not
included, so only one variable changed). Same instrument as the original 2%
reading — the Claude Code CLI statusline:

| step | window consumed |
| --- | --- |
| cold Obsidian start, no chat tab opened | **0%** |
| opening a new Claude tab | **0%** |
| selecting `opus` in the model picker, nothing sent | **0%** |

The plugin's own debug log confirms the mechanism rather than just the outcome
(timestamps UTC, machine is UTC+5):

```
14:52:28.403Z plugin              loaded
14:52:32.xxxZ provider.claude     modelCatalog.refresh.skipped  ← reason: seeded_on_first_use
14:54:40.xxxZ provider.claude     modelCatalog.refresh.skipped  ← reason: cache_fresh, ageMs 127716
14:57:02.679Z chat                send.requested                ← first real message, end of the clean window
```

No `refresh.started` / `.empty` / `.succeeded` for `claude` in the startup window
at all: the SDK session is not started, rather than started cheaply. Three
separate loads of the fixed build (10:34:09Z, 10:49:00Z, 14:52:32Z) all report
`seeded_on_first_use`.

Caveat worth stating: the CLI rounds to whole percent, so this proves "the 2% per
load is gone", not "the probe is now free".

## Scope: this does not remove the probe, only the per-load one

`refreshAttemptsByKey` still lives in process memory and the threshold is still
`10 * 60 * 1000`; the seed merely writes `lastAttemptAt = Date.now()`. Any picker
open (including the `ModelSelector` constructor on a new tab) more than ten
minutes after the last attempt still starts a full SDK session.

That is observed, not hypothetical — in the same log, on the pre-fix build:

```
09:07:52.237Z plugin              loaded
10:15:13.xxxZ provider.claude     modelCatalog.refresh.empty    ← real probe, ~67 min in, no reload
```

A longer-lived follow-up would be to persist the attempt timestamp across loads,
or to stop re-probing a fully `sdk`-sourced catalog on a timer and re-probe only
when the CLI/environment fingerprint changes.

## Not addressed here

`probeRuntimeCommands` is a second, independent SDK probe (slash-command/skill
discovery) with no TTL or seed cache at all — `ClaudeCommandCatalog` keeps only
in-session `sdkCommands`/`probePromise`. Its trigger is
`ClaudeCommandCatalog.listDropdownEntries` → `ensureProbed()` with an empty cache,
i.e. the first time the slash-command list is opened after a plugin load (chat
input `/`, the settings command list, or the inline-edit modal) — not tab creation
or activation: `claude` registers neither a `tabWarmupPolicy` nor a
`runtimeCommandLoader`, so its tab warmup mode resolves to `'none'` and
`prewarmProviderTab` returns immediately. Deliberately left out of this change.


sandsaber:
Thank you for this one — and for the way it is written.

The isolated measurement protocol, the log excerpts, and above all the "why it went unnoticed" section made this reviewable in minutes instead of an afternoon. Both blind spots you name are the interesting part of the bug: a reload test whose mock returned the same path at construction and at refresh, so it could never fail; and a plan indicator that reports `pctKnown: false` all day and therefore cannot distinguish 0% from 2%, on which an earlier investigation concluded "the probe is free". Diagnosing your own #85 rather than defending it is exactly the right instinct — as was #74/#80 on the Windows provider fixtures and #79 on the LF pin. It is appreciated.

I verified the diagnosis independently before touching anything:

- `ProviderWorkspaceRegistry.initializeAll` assigns `this.services[providerId]` only after `await …initialize()` (`ProviderWorkspaceRegistry.ts:51`), and `getCliResolver` reads that same map (`:108`). `getResolvedProviderCliPath` is therefore `null` for the entire construction window — `createClaudeModelCatalog` is called from the object literal `createClaudeWorkspaceServices` returns, strictly inside `initialize()`.
- Your new test fails on `main` (probe called once) and passes with the change. The regression is real.

Rather than send it back for another round, I pushed three commits to the branch.

## 1. `main` merged in

The branch was five commits behind (1.1.10, the plan-panel fixes, the `setImmediate` test fix), so its CI was measuring against a stale base. That also explains the "22 failed suites" baseline in the description — on the current base, macOS is fully green.

## 2. `fix(claude): keep the deferred seed from absorbing a config change`

The deferred seed is filed under the key computed at the first refresh. That is the right key for the CLI path — and it also picks up *current* settings. If the Claude configuration changed between the plugin load and that first refresh (environment variables, settings sources, Chrome), the seed lands on the new key and suppresses the probe that change should have triggered. The window is reachable: `GrimoireSettings.refreshProviderModelCatalog` runs when a provider is enabled, which happens before any tab's `ModelSelector` exists.

Only the CLI path was unknown at construction, so the seed now applies only while the rest of the key still matches the catalog that was persisted:

```ts
const seededCacheKey = buildClaudeModelCatalogCacheKey(
  initialSettings,
  plugin.getResolvedProviderCliPath?.('claude') ?? '',
);
if (seededCacheKey === cacheKey && currentSettings.discoveredModels.length > 0) {
  // seed + skip
}
```

New test: `still probes when the Claude config changed before the first refresh`. It fails without the guard.

## 3. `fix(providers): seed gemini, qwen, and codex catalogs when the CLI path resolves late`

Your root cause is not Claude-specific. `buildGeminiModelCatalogFingerprint`, `buildQwenModelCatalogFingerprint` and `buildCodexModelCatalogFingerprint` all contain `getResolvedProviderCliPath(…) ?? settings.cliPath`, and all three catalogs are constructed in the very object literal that creates their `cliResolver` — inside `initialize()`, same as Claude. Same `null`, so the seed was filed under the raw configured path while every refresh looked it up under the resolved one. Wherever those differ — auto-detection with no configured path, or a hostname-specific path — the seed missed and the ACP / app-server CLI warmed up on every plugin load. `opencode`, `mimocode` and `kimicode` fingerprint from settings alone and are unaffected.

The hold-back lives in `ProviderModelCatalogRefreshCache` so the three cannot drift apart: `seedOnFirstRefresh()` defers the seed, `applyDeferredSeed(fingerprint, hasCachedModels)` files it on the first refresh but only while nothing except the resolved path changed since construction. The fingerprint builders became pure (`settings`, `cliPath`, `environmentVariables`) so that deferred comparison can pin the environment text as it was at load — `getActiveEnvironmentVariables` reads live plugin state, so a closure over the initial settings alone would not have been enough. The live lookups moved into `resolve*ModelCatalogFingerprint()`. Same `seeded_on_first_use` debug event you introduced for Claude.

Ten new tests in total; the four that pin behaviour (`suppresses the reload warmup…` for gemini/qwen, `suppresses the reload listing…` for codex, and the Claude guard above) each fail without their fix.

Full local gate on the merged branch (macOS): **407 suites / 7144 tests passing**, `typecheck`, `lint` and `build:release` clean.

## Two things deliberately left alone

- **The eager seed branch.** In production it is unreachable — the resolver is never available at construction on a cold load — so the "a genuine mid-session CLI path change still forces a re-probe" property actually comes from the deferred seed's key, not from the eager one. Keeping it is still right: it is what the existing `still probes after a reload when the resolved CLI path changed` test exercises, and it stays correct if the construction order ever changes.
- **The 10-minute in-memory TTL.** Your scope note is accurate and your `10:15:13Z` log line proves it — a picker opened an hour into a session still starts a full SDK session with no reload involved. Persisting the attempt timestamp across loads, or re-probing an all-`sdk` catalog only on an environment-fingerprint change, deserves its own issue rather than being folded in here.

Merging once CI is green on the updated base.

PR #95, opened 2026-08-26
fix(grok): expose xhigh reasoning effort for grok-4.6

robintuffeau:
## Problem

Selecting Grok 4.6 in the Grimoire chat composer only offers **Low / Medium / High**. Grok Build and grok-4.6 also support **xhigh** (Extra high / Très élevé). That value already has i18n strings and is already accepted on the Grok launch path (`--reasoning-effort xhigh`); it just never appears in the picker for catalog models such as `grok-4.6`.

## Root Cause

`getSupportedThinkingOptionsForModel` only used the four-level native list (`GROK_NATIVE_THINKING_OPTIONS`, including `xhigh`) when `isGrokNativeBuildModelId()` matched. That helper is limited to `grok-build` and `grok-composer-*`.

The live catalog default is `grok-4.6` (and `grok-4.5`). Those ids fell through to `GROK_FALLBACK_THINKING_OPTIONS` (`low` / `medium` / `high`) whenever ACP had not yet discovered per-model thought levels — which is the usual case (`thinkingOptionsByModel` is empty).

Launch already forwarded `xhigh` via `normalizeGrokLaunchReasoningEffort`. The gap was the composer menu.

## Approach

Treat native xAI model ids (`grok-*` without a provider slash) as using the native thinking list, including `xhigh`. Third-party catalog models (`anthropic/...`, `minimax-token-plan/...`, …) still use the three-level fallback until ACP discovery fills `thinkingOptionsByModel`.

xAI documents that grok-4.6 accepts `xhigh` natively, and that grok-4.5 treats `xhigh` as `high`. That matches Grok Build CLI’s own fallback effort menu (`low` / `medium` / `high` / `xhigh`).

## Architecture

Architecture notes:

Grok-owned model identity and chat UI config only. No shared contract change. Launch already accepted `xhigh`; this PR only widens the picker for native Grok catalog ids.

## Security And Privacy

Security notes:

No security boundary change. Effort is still a user-selected launch argument already allowlisted by `GROK_LAUNCH_REASONING_EFFORT_VALUES`.

## Verification

Automated commands run:

```text
npm run test -- --selectProjects unit
npx eslint src/providers/grok/models.ts src/providers/grok/ui/GrokChatUIConfig.ts tests/unit/providers/grok/models.test.ts --max-warnings=0
npx tsc --noEmit
```

Unit: 408 suites, 7167 tests passed (plus the new grok-4.6 picker cases).

Manual verification:

Not run in Obsidian. Reproduced from a vault using `grok:grok-4.6` with empty `thinkingOptionsByModel`, which is the path this PR covers.

## Generated Artifacts

`npm run build:release` was not run; this is a small provider UI change with no CSS, dependency, or generated-artifact updates.

## Review Checklist

- [x] Protocol tests use real structured payloads and error codes.

### DIFF
--- a/src/providers/grok/models.ts
+++ b/src/providers/grok/models.ts
@@ -76,23 +76,18 @@ export function normalizeGrokLaunchReasoningEffort(
     : null;
 }
 
-export function isGrokNativeBuildModelId(rawModelId: string): boolean {
-  const normalized = rawModelId.trim();
-  if (!normalized) {
-    return false;
-  }
-
-  return normalized === 'grok-build' || normalized.startsWith('grok-composer-');
-}
-
-/** Native xAI model ids, as opposed to third-party catalog ids such as `anthropic/...`. */
+/**
+ * Native xAI model ids, as opposed to third-party catalog ids such as
+ * `anthropic/...`. Matched case-insensitively, like every other prefix test in
+ * this file, so a catalog that reports `Grok-4.6` is still recognised.
+ */
 export function isGrokNativeModelId(rawModelId: string): boolean {
-  const normalized = rawModelId.trim();
+  const normalized = rawModelId.trim().toLowerCase();
   if (!normalized || normalized.includes('/')) {
     return false;
   }
 
-  return normalized === 'grok' || normalized.startsWith('grok-');
+  return normalized === GROK_SYNTHETIC_MODEL_ID || normalized.startsWith('grok-');
 }
 
 export function encodeGrokModelId(rawModelId: string): string {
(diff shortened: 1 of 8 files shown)


sandsaber:
Thanks a lot for the PR! I changed the logic a little and made it dynamic per model, with the future in mind — in case the next models end up with different reasoning levels.

Grok Build reports the supported levels itself, in each model's `_meta`, and it does so for every available model rather than only the current one. Verified against a live `grok agent stdio` session (Grok Build 1.0.5) — both `session/new` and `session/load` carry it:

```
grok-4.6: ["xhigh", "high", "medium", "low"]
grok-4.5: ["high", "medium", "low"]
```

Previously `normalizeGrokAcpSessionModels` rebuilt each model from `id`/`name`/`description` and dropped `_meta`, so that answer never reached the picker and the supported levels had to be derived from the model id. Now `thinkingOptionsByModel` is seeded from what the session reports, so a single session makes the picker exact for every model, and a new model arrives with its own set without a code change.

Your static list stays as the fallback until the first session, and `xhigh` is kept in it — new models will most likely support it.

Precedence is unchanged where the two sources overlap: `thought_level` still wins for the active model. The one adjustment is that an absent option no longer deletes the active model's levels when the session did report them — `session/new` carries the per-model levels and no `thought_level` option at all, so the old `delete` branch discarded exactly what had just been learnt.

PR #96, opened 2026-08-26
feat(antigravity): stream answer text and tool steps while the run is open

greenkms:
## Problem

The Antigravity provider awaits `runPrint` and yields the whole answer only after
the child closes, so a multi-minute turn shows nothing but a spinner. The parser
already sees every frame — it just drops everything that is not the final
`result`.

The frames are there and their meaning is already documented in this codebase:
`formatAntigravityAnswerWithNotice` notes that "the `text_delta` frames sum
exactly to `result.response`". Today that fact is only used to justify keeping an
answer when agy flags ERROR after answering (#69); nothing renders the deltas as
they arrive.

## Change

Decode `step_update` frames into progress events and drive the turn from a queue
the callbacks fill and the generator drains, so the answer and the tool calls
appear as they happen.

- **Text.** `text_delta` frames stream straight into the chat. Because the deltas
  sum to `result.response` exactly, only the remainder the final answer adds on
  top of them is emitted at the end. That remainder keeps its leading blank line,
  which is what separates the trailing `> Warning:` note agy appends after an
  ERROR-flagged success from the answer it follows — so #69's behaviour is
  preserved rather than re-derived.
- **Tools.** Tool steps become native `tool_use`/`tool_result` chunks keyed by
  `step_index`, so they render as real tool cards instead of text lines. The
  `ACTIVE` frame carries the arguments in `tool_info.parameters` and the `DONE`
  frame carries what the tool printed in `tool_info.output`, so a card opens with
  the call and closes with its output.
- **Summaries.** agy names its tool parameters `CommandLine`, `AbsolutePath` and
  friends. None of those were recognised by the tool-summary helper, so an agy
  card showed a bare tool name. They are added to the recognised list, and the
  lookup is split out from the helper's first-value fallback so that tools which
  deliberately render no summary (`AskUserQuestion`) keep getting an empty string
  rather than a JSON dump of their input. MCP tools keep the fallback they have
  always had.
- **Liveness.** The log-file growth already polled for liveness (#73) now also
  emits a status tick, so a pipe-silent tool call stops looking like a hang. No
  second timer is introduced.

Nothing is streamed on the transcript-recovery and legacy `--print` paths, so
those still deliver the answer whole.

Copy added to all ten locales.

## How the frame shape was established

The tool frames were read off a live capture rather than inferred, because an
earlier draft of this change assumed agy never streams tool output and closed
every card with an empty result. Running

```
agy --dangerously-skip-permissions --input-format stream-json --output-format stream-json
```

with a one-command prompt produces, on the `DONE` frame:

```json
{"step_index":2,"state":"DONE","step_type":"tool","tool_name":"run_command",
 "duration_seconds":0.3985843,
 "tool_info":{"name":"run_command",
              "parameters":{"CommandLine":"echo grimoire-capture-test"},
              "output":"grimoire-capture-test\r\n"}}
```

A missing or non-string `output` still yields an empty string, so frames without
the field and tools that genuinely print nothing behave as before. The test
fixtures are copied verbatim from that capture.

## Scope

Only the streaming `--print` path changes, plus the tool-summary helper described
above. Cancellation, the add-dir probe, the five-minute timeout work and the
ERROR-with-answer handling (#67, #69, #70, #73) are untouched and their tests are
unmodified.

## Tests

Eighteen cases added — nine to `AntigravityStreamJson`, six to
`AntigravityChatRuntime`, three to `ToolCallRenderer`:

```
AntigravityStreamJson
  reports text deltas in arrival order while the run is still open
  pairs a tool start with its completion through the shared step index
  reports what the tool printed from the completion frame
  treats a non-string output as no output rather than rendering it
  keeps sequential tools apart when agy omits the step index
  ignores lifecycle frames that carry no user-visible progress
  concatenates the streamed deltas into exactly the final response
  still parses the result when no progress listener is attached
  survives a listener that throws so one bad frame cannot abort the run

AntigravityChatRuntime
  streams text deltas into the chat before agy exits
  renders agy tool steps as tool cards while the run is open
  closes a tool card with the output agy reported for the step
  appends only the untold tail when the result adds a warning to a streamed answer
  still delivers a recovered transcript answer that was never streamed
  reports an empty response as an error even after tool cards were rendered

ToolCallRenderer
  returns the command line for an antigravity run_command step
  returns the target path for antigravity file tools
  does not invent a summary from an unrecognised parameter name
```

The last two `AntigravityChatRuntime` cases pin the paths that must *not* change:
transcript recovery still delivers a whole answer, and an empty response is still
an error even once tool cards have been drawn.

Full local gate on Windows: 22 failed suites / 35 failed tests both before and
after — an exact match with this machine's `main` baseline at 2db77cc, compared
as a list of failing suites and not only as a total; passing tests 7328 → 7346,
which is exactly the eighteen cases added. `typecheck` and `lint` are clean.

The failing suites are pre-existing Windows-environment failures on this machine,
unrelated to this change. No existing test was modified.

### DIFF
--- a/src/features/chat/controllers/StreamController.ts
+++ b/src/features/chat/controllers/StreamController.ts
@@ -1893,6 +1893,18 @@ export class StreamController {
 
     // If indicator already exists, just re-append it to the bottom
     if (state.thinkingEl) {
+      // An override names what is actually happening right now, so it has to
+      // reach an indicator that is already up - which is the usual case once a
+      // turn is running. Only the label is replaced; the elapsed-time hint that
+      // follows it keeps counting.
+      if (overrideText) {
+        // The label is the first span; the elapsed-time hint that follows it
+        // keeps counting and must not be replaced.
+        const labelEl = state.thinkingEl.children[0] as HTMLElement | undefined;
+        if (typeof labelEl?.setText === 'function') {
+          labelEl.setText(overrideText);
+        }
+      }
       state.currentContentEl.appendChild(state.thinkingEl);
       this.deps.updateQueueIndicator();
       return;
--- a/src/features/chat/rendering/ToolCallRenderer.ts
+++ b/src/features/chat/rendering/ToolCallRenderer.ts
@@ -86,12 +86,28 @@ function getMcpOperationName(name: string): string {
   return operation.replace(/^obsidian_/, '');
 }
 
-function getDefaultToolSummary(input: Record<string, unknown>): string {
-  const preferredKeys = ['query', 'pattern', 'dirpath', 'filepath', 'file_path', 'path', 'url', 'command'];
-  for (const key of preferredKeys) {
+const PREFERRED_SUMMARY_KEYS = [
+  'query', 'pattern', 'dirpath', 'filepath', 'file_path', 'path', 'url', 'command',
+  // Antigravity (agy) names its tool parameters in PascalCase.
+  'CommandLine', 'AbsolutePath', 'TargetFile', 'SearchPath', 'SearchDirectory', 'DirectoryPath',
+];
+
+/**
+ * Summary drawn only from parameter names we recognise. Tools that deliberately
+ * render no summary (AskUserQuestion and friends) must keep getting '' here, so
+ * this never falls back to dumping an arbitrary value.
+ */
+function getPreferredToolSummary(input: Record<string, unknown>): string {
+  for (const key of PREFERRED_SUMMARY_KEYS) {
     const value = getInputText(input, key);
     if (value) return truncateText(value, 60);
   }
+  return '';
+}
+
+function getDefaultToolSummary(input: Record<string, unknown>): string {
+  const preferred = getPreferredToolSummary(input);
+  if (preferred) return preferred;
 
   const firstValue = Object.values(input).find(value => stringifyToolValue(value));
   return firstValue ? truncateText(stringifyToolValue(firstValue), 60) : '';
@@ -203,7 +219,7 @@ export function getToolSummary(name: string, input: Record<string, unknown>): st
       if (isAgentLifecycleTool(name)) {
         return getAgentLifecycleSummary(name, input);
       }
-      return '';
+      return getPreferredToolSummary(input);
   }
 }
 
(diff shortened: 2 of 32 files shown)


sandsaber:
Thanks for the PR! Great catch that the parser already sees every frame and only the `result` was being used — the live streaming is a real improvement. I merged `main` in and pushed some fixes on top; here is what changed and why.

**Correctness, in the order a run hits them:**

- `text_delta` was emitted for every step, including a tool step, because the emit sits above the `step_type !== 'tool'` guard. Only a non-tool step narrates the answer — text that `result.response` never contains breaks the prefix your untold-tail logic depends on.
- A frame without `step_index` was counted from zero into the same namespace as real indices, so a mix of indexed and un-indexed frames merged two different calls onto one card. Ids now carry their origin (`step-N` vs `unindexed-N`).
- A completion whose `ACTIVE` half never arrived was dropped, taking the call and its output out of the transcript. It names its tool and carries its output, so the card is now opened from the completion.
- The parser was flushed *after* `settled` closed the progress channel, so a trailing frame written without its newline was parsed and then discarded — a tool `DONE` sitting there left its card open. The flush now happens first, which is the order `expire()` already used.
- A run that ended between `ACTIVE` and `DONE` (cancelled, timed out, ERROR mid-tool) left its card spinning, and nothing downstream reconciles a tool call left running — it was persisted that way. Whatever is still open is now closed as the turn ends.
- The tail used an exact `startsWith` test, so any divergence between the deltas and `result.response` re-emitted the whole answer the user had just watched arrive. Only the part past the common prefix is appended now.
- Streaming dropped the `output.trim()` the whole-answer path applied, so agy's leading blank lines reached both the bubble and the saved message.
- The localized liveness string never actually rendered: `showThinkingIndicator` returns early when an indicator is already up — the usual case mid-turn — without applying the override, so the seconds froze at their first value.

**Provider boundary:**

The agy PascalCase argument names were taught to `ToolCallRenderer`, which is shared chat rendering — that changed tool cards for every provider, and agy's own cards stayed degraded anyway (fallback icon, raw tool name, no diff rendering) because `run_command` was never mapped to `Bash`. I reverted the shared file and moved the vocabulary into a provider-owned normalization module, the same shape `codex`, `grok`, `kimicode`, `mimocode` and `opencode` already use. Native argument names are kept beside the neutral ones so a card still lines up with agy's own logs. Only tools whose native name is confirmed against the CLI are mapped; the argument aliases apply to every agy tool, so one the table does not know still shows what it is acting on.

**Left alone, deliberately:**

- The heartbeat and the provider-neutral silent-turn indicator take turns on screen roughly every 10–15s, because every chunk resets the silence timer. Reconciling two indicators is a change to shared turn feedback rather than to this feature.
- The queue in the generator is a seventh copy of a class six provider runtimes already carry privately. Extracting it touches all of them, so it does not belong in this PR.

Local gate: 7218 unit tests, typecheck and lint all clean.

Issue #93, opened 2026-08-25
The catalog seed adopts a CLI change made between sessions instead of detecting it


greenkms:
## Summary

#92 removed the timer, and that turned an assumption the seed had always been
making into a permanent one. The seed files the persisted catalog under whatever
key *this load* computes; nothing records the key the catalog actually came from.
So it cannot distinguish "discovered under this configuration" from "assumed to
match it".

While the TTL existed, a wrong assumption expired within ten minutes. It no longer
expires. A CLI swapped while Obsidian was closed is therefore adopted rather than
detected, and the previous CLI's model list stays pinned indefinitely.

## What happens now

Close Obsidian, change the CLI, reopen. Construction and lookup both read the new
CLI, so the key they compare is self-consistent: the seed applies, the probe is
suppressed, and nothing will ever re-check.

`claudeCliBinaryFingerprint` cannot cover this case. It only helps when the key
differs from one recorded earlier *in the same process*; across a reload nothing
was recorded. The result runs opposite to the intent stated in #92 ("without a
timer, an upgrade in place would otherwise never surface new models"):

| CLI upgraded | detected? |
| --- | --- |
| while Obsidian is running | yes — the key changes against the in-memory attempt |
| while Obsidian is closed | no — the seed adopts the new key, permanently |

`re-probes when the CLI binary changes behind an unchanged path` passes because it
flips the fingerprint *after* the catalog is constructed, which is the first row
only.

## Evidence

Two cases added to `tests/unit/providers/claude/app/ClaudeModelCatalog.test.ts`,
both failing on `main` at 2db77cc (probe calls: expected 1, received 0):

```ts
it('re-probes when the CLI binary changed while the plugin was not running', async () => {
  const settings: Record<string, unknown> = {};
  updateClaudeProviderSettings(settings, {
    enabled: true,
    discoveredModels: [{ id: 'opus', displayName: 'Opus', source: 'sdk' }],
  });
  mockedProbe.mockResolvedValue([{ id: 'sonnet', displayName: 'Sonnet', source: 'sdk' }]);

  // The new binary is already in place when the catalog is constructed.
  mockedBinaryFingerprint.mockReturnValue('120:2');
  const plugin = createPlugin(settings);
  await createClaudeModelCatalog(plugin).refreshModels({ plugin, settings });

  expect(mockedProbe).toHaveBeenCalledTimes(1);
});
```

The same shape with `getResolvedProviderCliPath` returning `/opt/claude` from
[...]

PR #97, opened 2026-08-26
fix(claude): stop paying for a command probe on every plugin load

greenkms:
## The problem

Opening the slash-command dropdown for the first time after a plugin load starts a
full Claude Code session and bills the plan window for it. There is no persistence,
so every Obsidian restart pays again.

`ClaudeCommandCatalog.listDropdownEntries()` probes when `sdkCommands.length === 0`.
After a successful probe the list is non-empty, so within one load the probe runs
exactly once - and that single probe is what the measurement below caught. Nothing
survives the load, so the next start repeats it.

There is a second, worse mode. After an *empty* probe (no CLI installed, or a
session that is not authenticated) `sdkCommands` stays at zero, and there is no
retry window at all, so the probe fires on **every single dropdown open**.

And a third path, which is easy to miss: `TabManager.ts:1037` calls
`setRuntimeCommands([])` for a blank tab that skips warmup. A live code path
routinely returns the catalog to the "probe needed" state, not just a cold start.

## The measurement

Taken with the plugin untouched (`main.js` MD5 `0B755CA12987CBBCA168D7F1B604DC72`),
by an agent running on a different provider, with no other agentic Claude session
sharing the five-hour window:

| Point | What was done | Window |
|---|---|---:|
| A | before Obsidian launch | 72 % |
| B | loaded, idle 60 s, no Claude tab | 72 % |
| C | new Claude tab opened, nothing sent, 30 s | 72 % |
| D | slash-command list opened once with `/`, nothing sent, 30 s | **73 %** |

`B − A = 0`, `C − B = 0`, `D − C = 1` percentage point. The step is isolated to the
first dropdown open.

This is not a claim that the probe costs "exactly 1 %": the statusline rounds to
whole percent, so the true price is bounded, not measured. What is established is a
billed transition attributable to the probe alone.

Worth stating plainly: `probeRuntimeCommands` carried the comment "no API call, no
cost". That comment was wrong, and this PR corrects it.

## The fix

Two layers, one defect in two modes.

**1. Persist the list together with the key it came from.** New settings fields
`discoveredCommands` and `discoveredCommandsFingerprint`. On the first dropdown open
the catalog restores the list when the recorded digest still matches the current
configuration, and probes only when there is nothing to reuse.

The digest covers the inputs that decide what the SDK can see at all: CLI binary
fingerprint, CLI path, `enableChrome`, environment hash, environment variables,
`loadUserSettings`, project-settings env hash, `respectProjectSettings`. Only the
FNV-1a digest is persisted - the key embeds raw environment variables, which can
hold an API key.

The fingerprint is computed lazily, at first use rather than in the constructor.
That structurally avoids the deferred-seed dance the model catalog needs: there,
the seed had to be postponed because `getResolvedProviderCliPath` still returns
`null` while `ProviderWorkspaceRegistry.initialize()` is running. This is the
lesson of #93 applied directly - record the key the catalog *came from*, not the
one the current load happens to compute.

**2. Pace retries after an empty probe.** A ten-minute in-memory window per
fingerprint, the same value as `EMPTY_ATTEMPT_RETRY_MS` in the model catalog. In
memory on purpose: persisting it would keep the dropdown empty across the very
restart meant to fix a missing CLI. A probe that *throws* is treated the same way -
the cause is usually a broken or absent CLI, and retrying it per keystroke is what
the window exists to stop.

## Why this is not the model-catalog fix copied over

The SDK command list includes vault commands and skills: the SDK scans
`.claude/commands/` and `.claude/skills/` itself. Models change every few weeks;
commands change every time the user writes a skill. Keying the cache on folder
mtimes would therefore make **editing a skill cost a probe** - the opposite of the
goal.

Instead the cache is never invalidated by vault contents. A list restored from the
cache is merged with `listVaultEntries()`, which is a plain file read and costs
nothing. On a name collision the vault entry wins: it is a real, editable file the
user owns, whereas the cached entry is only a description of it. A list handed over
by a live session is authoritative and is shown byte-for-byte as before - the merge
applies only to a restored snapshot, so the diff stays inside the bug.

`setRuntimeCommands` now also refreshes the cache for free whenever a live session
reports commands, and deliberately does *not* clear it on the empty-list reset, so
the blank-tab path above is covered too.

## Verifiability

The probe was not instrumented, so the measurement's own log check could only prove
the *absence* of other activity, never the probe itself. It is instrumented now:
`commandCatalog.probe.started|succeeded|empty|failed|skipped` under scope
`provider.claude`, with `skipped` carrying `cache_fresh` or
`empty_attempt_throttled`. The effect of this patch is readable straight from
`.grimoire/logs/*.jsonl`.

## Notes for review

- `ClaudeModelCatalog.ts` is deliberately untouched: it is in flight as #94, and
  editing it here would guarantee a conflict. The cache-key builder and the FNV-1a
  helper are therefore duplicated rather than extracted. This is a considered
  choice, not an oversight - the two keys can legitimately diverge, since the
  command list depends on vault scanning that models do not care about.
- `ProviderCommandCatalog.refresh()` for Claude was a no-op while already being
  wired to the refresh button in `ProviderSkillSettings.ts:185`. It now clears the
  cache and forces one probe. That makes it the only UI element that knowingly
  spends a probe - the escape hatch, opt-in by the user.
- The catalog receives the cache and the log sink as an optional fourth `deps`
  argument, so it holds no reference to the plugin and is tested against a fake.
  Constructed without `deps` it behaves exactly as it does today, which is covered
  by a regression test.

## Testing

- `npm run test -- --selectProjects unit` - 7140 passed. 35 failures in 22 suites
  are pre-existing on `main`: verified by running the same suite on a pristine
  `main` worktree, which fails identically (same 22 suites, same 35 tests). None of
  them touch the files in this PR.
- `npm run typecheck`, `npm run lint`, `npm run build:release` - all clean.
- 28 new tests: cold-start probes once; no probe on a matching fingerprint; probe
  and rewrite on a diverged one; merge with the vault and vault-wins-collision; the
  live-session path unchanged; the empty reset preserving the cache; the empty-probe
  window holding and then expiring; a thrown probe throttled; `refresh()` forcing a
  probe through both cache and window; a rejected cache write not breaking the
  dropdown; the no-deps regression guard; and, for the store, that no raw
  environment value is ever persisted.


sandsaber:
Thanks for the PR — the measurement in the description is what makes this convincing, and correcting the "no API call, no cost" comment was overdue. I merged `main` in (it now carries #94's model-catalog fingerprint, which conflicted additively with the fields here) and pushed fixes on top.

**Three things found against `main`:**

- `setRuntimeCommands` filed the cache whenever a live session handed over a non-empty list, with no check that anything had changed — and `TabManager.getSdkCommands` calls it from the dropdown's own entry provider, from tab prewarm and from runtime-ready. Opening the slash menu rewrote `grimoire-settings.json` with byte-identical content, which is a file the user's vault sync sees. The store now compares against what is persisted and returns without saving when nothing moved.
- `refresh()` had no way in. The only caller is `ProviderSkillSettings`, which is wired for grok, opencode, mimocode, qwen and kimicode — not for Claude, whose settings tab renders its own skills UI. (The PR description's `ProviderSkillSettings.ts:185` claim points at a component Claude never constructs.) There is now a **Refresh commands** button in the slash-commands section, mirroring Refresh models.
- `buildCommandCatalogCacheKey` and its FNV digest were byte-identical to the pair #94 added for the model catalog, comment included. Both now use one `claudeCatalogCache` module — along with the retry window, whose two literals each carried a comment claiming to match the other.

**Then from a review pass:**

- **A list could be filed under a configuration it never came from.** `writeCache` read the digest at write time instead of carrying the one the discovery ran under, so a settings edit or CLI upgrade landing mid-probe stamped the old configuration's commands with the new key. With no expiry that never heals. The digest now travels with the list, and a write whose configuration moved underneath it is skipped.
- **Refresh destroyed the list before knowing it could replace it.** It cleared the cache, then probed — so a refresh against a broken or logged-out CLI lost a working list, dropped the dropdown to vault entries only, then throttled every retry for ten minutes. It now discovers first and keeps what it had when discovery returns empty. This matters more now that the button can actually be reached.
- **A forced refresh could be spent and thrown away.** It yielded on the cache write, and a session handing over commands in that gap made the probe's result non-applicable — a billed session discarded while the settings tab reported success. A forced probe now wins, and one already in flight is awaited rather than joined by a second.
- **No digest meant no pacing at all.** When `currentFingerprint()` threw, nothing was recorded and nothing skipped — restoring the probe-per-keystroke this class exists to prevent, since the dropdown reopens on every input event. A null digest now collapses to one shared bucket.
- **The failure that matters was invisible.** `probeRuntimeCommands` swallowed every SDK error and returned `[]`, so a missing or unauthenticated CLI logged as `probe.empty` with no detail and `probe.failed` was unreachable. The abort that ends a successful probe is still ignored; anything else is reported.
- **A command and a skill sharing a name collapsed into one.** Both are legal on disk and separately editable, and the non-merged branch returns both, so which survived depended on whether the runtime list came from cache. Vault entries are keyed by kind as well as name now.
- Smaller: a rejected settings save no longer leaves memory claiming a record disk never took, in either direction; persisted commands are trimmed and deduplicated like the models beside them; the digest is computed once per dropdown open instead of three times, each of which stats the CLI binary; and the `Date.now` stubs in the throttle tests are restored in `afterEach` rather than as the last statement of a body an assertion can throw out of.

**Left alone, deliberately:** the merged list still rescans the vault on every open. That scan is what keeps a skill created since the cache was written visible, so memoizing it would trade away the freshness the merge exists for.

Local gate: 7257 unit tests, typecheck and lint all clean.

Issue #99, opened 2026-08-27
A failed session resume re-sends the whole transcript: five runtimes still carry the pattern


greenkms:
Filed alongside the MiMoCode fix (PR: fix/mimocode-session-history-replay).
That PR deliberately touches one provider - the one the behaviour was
reproduced and measured on. The shape of the defect is not MiMoCode-specific,
so this is the part that outlives it.

## What happens

A conversation with a persisted session is reopened. `session/load` fails.
Grimoire opens a fresh session and hands it the entire prior conversation as
prompt text. On the measured case that first message cost roughly 8.4M
plan-usage credits (MiMoCode's credit unit; ~34k tokens) against ~71k credits
for the same conversation once running - for a short question that came back
in ten seconds with no tool calls.

Nothing tells the user this happened. The turn simply costs what an entire
transcript costs, once per failed resume.

## The mechanism, in two halves

**1. The session id is gone before `query()` can tell why.**

```ts
const expectedSessionId = this.sessionId;
let shouldBootstrapHistory = previousMessages.length > 0
  && (!expectedSessionId || this.sessionInvalidated);
...
if (expectedSessionId && !this.sessionId) {
  shouldBootstrapHistory = previousMessages.length > 0;
}
```

Both branches lead to a replay. The second one is unconditional: if readiness
dropped the session for any reason, the transcript goes out.

**2. Tab warmup drops it even earlier.**

`ProviderRuntimeCommandLoader.loadCommands()` reuses the *bound* tab runtime
when the conversation already has a session id, and calls `ensureReady()` on
it before the user has typed anything. A `session/load` failure there clears
`sessionId` while the composer is still empty. `query()` then starts with no
session id at all - which is indistinguishable, at that point, from a cold
resume that never had one. So even a fix that reserves bootstrap for
`!expectedSessionId` still replays on the single path an editor restart takes.

This is what makes the defect hard to see in review: the guard reads correctly
and is still wrong, because the state it inspects was already destroyed by a
different call site.

## Where it still lives (line numbers on `main` @ 0d57617)

| Runtime | `shouldBootstrapHistory` init | unconditional re-set |
|---|---|---|
| `src/providers/kimicode/runtime/KimicodeChatRuntime.ts` | 391 | 409 |
| `src/providers/opencode/runtime/OpencodeChatRuntime.ts` | 391 | 409 |
| `src/providers/gemini/runtime/GeminiChatRuntime.ts` | 262 | 279 |
[...]

PR #100, opened 2026-08-27
fix(mimocode): stop replaying the whole transcript when a session fails to resume

greenkms:
## Problem

The first message sent after an Obsidian restart, in a MiMoCode conversation
that had a persisted session, cost roughly 8.4M plan-usage credits against
~71k in the same conversation once it was running. Same short question, ten
seconds, zero tool calls.

What was being paid for was the transcript: `session/load` fails, Grimoire
opens a fresh session, and the fresh session was handed the entire prior
conversation as prompt text. Nothing tells the user this happened - the turn
simply costs what a whole transcript costs, once per failed resume.

The 8.4M figure is plan-usage **credits**, not tokens; MiMoCode's credit unit
is price-weighted and runs a few hundred times larger. The same first message
measures ~34k tokens. The unit does not change what the code was doing, but it
is worth naming so the magnitude is not mistaken for a token-accounting bug.

Fixes #

## Root Cause

Three defects in sequence.

**1. `loadSession()` only absorbed one error shape.** The catch narrowed on
`isAcpMissingSessionError`, so an auth failure, a database error or any other
JSON-RPC error propagated as an exception and left the stale `sessionId` in
place for the next turn to trip over.

**2. `query()` replayed unconditionally when readiness dropped the session.**

```ts
const expectedSessionId = this.sessionId;
let shouldBootstrapHistory = previousMessages.length > 0
  && (!expectedSessionId || this.sessionInvalidated);
...
if (expectedSessionId && !this.sessionId) {
  shouldBootstrapHistory = previousMessages.length > 0;   // unconditional
}
```

Both branches lead to a replay, and the second ignores why the session went
away.

**3. The session is usually gone before `query()` even starts.** This is the
part that makes the defect survive an obvious-looking fix.
`mimocodeTabWarmupPolicy.resolveMode()` always returns `'commands'`, so tab
warmup goes through `MimocodeRuntimeCommandLoader`. When the conversation
already has a session id, `canReuseRuntime` is true and the loader calls
`ensureReady()` on the **bound tab runtime** - before the user has typed
anything. A `session/load` failure there clears `sessionId` while the composer
is still empty. `query()` then starts with no session id at all, which at that
point is indistinguishable from a cold resume that never had one. So even
reserving bootstrap for `!expectedSessionId` still replays on the single path
an editor restart takes.

Protocol evidence from the plugin's debug log: `session.load_failed` carrying
`JsonRpcErrorResponse` code `-32603`, twice on the same session id, followed by
a 9.7s turn with zero tool calls.

## Approach

Three commits, one per defect above.

`81b481c` - the error class decides: a `JsonRpcErrorResponse` from
`session/load` clears the binding (`sessionId = null`,
`sessionInvalidated = true`) and returns false. Transport-level errors (closed
pipe, failed spawn) still propagate, because those say nothing about whether
the session is usable.

`05d3ea2` - the unconditional re-set in `query()` is removed. History bootstrap
is reserved for a cold resume.

`709cbca` - bootstrap additionally requires `!this.sessionInvalidated`.
That flag already means "we had a session and lost it", and it is the only
state that survives the warmup path with the answer in it. The previous code
used the same flag with the opposite sign, as a *reason* to replay.

Alternative considered and rejected: keeping the replay but asking first. That
is a better end state and needs a UI decision, so it is raised in the linked
issue rather than smuggled in here. The behaviour this PR ships is the
conservative half - a session dropped by a load failure starts clean, and
recovering the earlier context is the user's to do. Silently re-buying the
whole transcript is not a recovery anyone consented to.

Deliberately unchanged: the mid-turn retry after a transport close still
bootstraps when it lands in a different session, because that is one
interrupted turn being finished rather than a new conversation being invented.
A genuine cold resume still gets its history, and that case now has a test of
its own so the new guard cannot quietly swallow it.

## Architecture

Architecture notes:

Both changed files are under `src/providers/mimocode/`. No shared contract is
touched: `JsonRpcErrorResponse` and `isAcpMissingSessionError` are used as they
already exist in `src/providers/acp/`, and no signature changes.

Sibling check was run and the result is not clean.
`shouldBootstrapHistory` is computed identically in `KimicodeChatRuntime`
(391/409), `OpencodeChatRuntime` (391/409), `GeminiChatRuntime` (262/279),
`GrokChatRuntime` (521/542) and `QwenChatRuntime` (285/302), and
`KimicodeRuntimeCommandLoader` / `OpencodeRuntimeCommandLoader` are
line-for-line analogues of the MiMoCode loader, so the warmup half of the
mechanism applies to them as written. This PR stays on the one provider the
behaviour was reproduced and measured on; the class is filed as an issue rather
than fixed blind across five runtimes with their own session lifecycles and
their own tests.

## Security And Privacy

Security notes:

No security boundary change. The failure behaviour narrows rather than widens:
a session that cannot be resumed is dropped and replaced instead of being kept
around in a stale state.

Worth naming as a privacy improvement rather than a neutral one. Prior to this
change, a resume failure silently re-sent the full contents of the
conversation - which in Obsidian routinely includes note text pulled in as
context - to a newly created remote session, with no prompt and no indication.
That now does not happen unless the conversation genuinely never had a session.

Added tests assert on locally constructed strings only; no fixtures, paths or
provider payloads were added.

## Verification

Automated commands run:

```text
npm run typecheck
npm run lint
npm run build:release
npx jest --testPathPatterns "MimocodeChatRuntime"
```

to load and to open the view.

`tests/unit/providers/mimocode/MimocodeChatRuntime.test.ts`: 43 passed, 2
failed. The `main` baseline of the same file on the same machine is 40 passed,
2 failed - the two failures are the pre-existing Windows path-separator
assertions in `resolveSessionPath` (`/tmp` vs `D:\tmp`), unchanged by this
branch and measured in a scratch worktree at `0d57617`.

Three tests added: the readiness drop, the warmup drop, and the cold resume
that must still bootstrap.

One existing test changed meaning and is renamed, so it is not mistaken for
breakage: `preserves the saved session binding when session/load fails
transiently` -> `clears the session binding when session/load fails with any
RPC error`. That is the assertion for the first commit above, and it is the one
behavioural reversal in this PR: an RPC failure that used to keep the stale
binding and rethrow now clears it and returns false. Kimicode and Opencode
still carry the original assertion, deliberately - see the linked issue.

Full unit suite on Windows, this branch: 392 suites passed, 22 failed; 7422
tests passed, 36 failed, 8 skipped. The `main` baseline at `0d57617` on the
same machine: 392 passed, **the same 22** failed; 7419 passed, **the same 36**
failed, 8 skipped. The delta is exactly the three added tests.

Manual verification:

Windows 11, Obsidian, MiMoCode provider. The reproduction that produced the
8.4M figure and the `-32603` log entries was run against the affected build,
and the first two commits were run in the live vault afterwards. The third
commit is covered by automated tests only - reproducing it on demand requires
`session/load` to fail during warmup, which is not something the plugin can be
asked to do from the UI. Its failure mode was confirmed by test against the
installed build before the fix, not by hand afterwards.

## Generated Artifacts

No dependency change; `package.json` and `package-lock.json` are untouched. The
branch changes exactly two files.

## Review Checklist

- [x] Protocol tests use real structured payloads and error codes.

Issue #99

### DIFF
--- a/src/providers/mimocode/runtime/MimocodeChatRuntime.ts
+++ b/src/providers/mimocode/runtime/MimocodeChatRuntime.ts
@@ -398,7 +398,17 @@ export class MimocodeChatRuntime implements ChatRuntime {
     // whole transcript (observed: ~8.4M plan-usage tokens on the first message
     // after an Obsidian restart, vs ~71k steady state). History bootstrap stays
     // reserved for a genuine cold resume where we never held a session id.
-    let shouldBootstrapHistory = previousMessages.length > 0 && !expectedSessionId;
+    //
+    // `sessionInvalidated` is what tells the two apart, because the session can
+    // be dropped before query() ever runs: tab warmup goes through the runtime
+    // command loader, which calls ensureReady() on the bound runtime, and a
+    // session/load RPC failure there clears sessionId and raises the flag.
+    // Without this guard query() would see a null session id, read it as a cold
+    // resume and replay the transcript anyway - which is the whole cost this is
+    // meant to remove, on the single path most likely to hit it.
+    let shouldBootstrapHistory = previousMessages.length > 0
+      && !expectedSessionId
+      && !this.sessionInvalidated;
 
     const lifecycleGeneration = this.lifecycleGeneration;
     if (!(await this.ensureReadyForQuery(lifecycleGeneration))) {
--- a/src/providers/mimocode/runtime/MimocodeChatRuntime.ts
+++ b/src/providers/mimocode/runtime/MimocodeChatRuntime.ts
@@ -393,8 +393,12 @@ export class MimocodeChatRuntime implements ChatRuntime {
   ): AsyncGenerator<StreamChunk> {
     const previousMessages = conversationHistory ?? [];
     const expectedSessionId = this.sessionId;
-    let shouldBootstrapHistory = previousMessages.length > 0
-      && (!expectedSessionId || this.sessionInvalidated);
+    // Owner decision: a lost or invalidated session must NOT auto-replay the
+    // conversation history into a fresh session. Doing so silently spends the
+    // whole transcript (observed: ~8.4M plan-usage tokens on the first message
+    // after an Obsidian restart, vs ~71k steady state). History bootstrap stays
+    // reserved for a genuine cold resume where we never held a session id.
+    let shouldBootstrapHistory = previousMessages.length > 0 && !expectedSessionId;
 
     const lifecycleGeneration = this.lifecycleGeneration;
     if (!(await this.ensureReadyForQuery(lifecycleGeneration))) {
@@ -410,9 +414,10 @@ export class MimocodeChatRuntime implements ChatRuntime {
     }
 
     const cwd = getVaultPath(this.plugin.app) ?? process.cwd();
-    if (expectedSessionId && !this.sessionId) {
-      shouldBootstrapHistory = previousMessages.length > 0;
-    }
+    // Intentionally NO history bootstrap here: if readiness dropped the session
+    // (session/load RPC failure, forced restart), the new session starts clean.
+    // The prior behaviour re-sent the entire transcript and burned the context
+    // window without the user asking. Recovery of prior context is manual.
 
     if (!this.sessionId) {
       const sessionId = await this.createSession(cwd);
--- a/src/providers/mimocode/runtime/MimocodeChatRuntime.ts
+++ b/src/providers/mimocode/runtime/MimocodeChatRuntime.ts
@@ -71,8 +71,8 @@ import {
   extractAcpSessionModelState,
   extractAcpSessionModeState,
   extractAcpSessionThoughtLevelState,
-  isAcpMissingSessionError,
   isAcpRetryableTransportClose,
+  JsonRpcErrorResponse,
   planAcpEnsureReadySessionPhase,
   resolveWorkspacePath,
   runAcpEnsureReadyForQuery,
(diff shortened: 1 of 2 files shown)


sandsaber:
You were right that the error class has to decide this, and right that the replay was the expensive part. I went to check what the CLIs actually say, and it turns out the ground is worse than either of us assumed.

I resumed a session id that no agent has ever seen:

- MiMoCode 0.1.13 — `-32603 "Internal error"`, `data: {}`
- OpenCode 1.18.18 — `-32603 "Internal error: OpenCode service failure"`
- Kimi Code 0.31.1 — `-32602 "Invalid params: Unknown sessionId: …"`

Run those through `isAcpMissingSessionError` and all three come back false. The first two carry nothing to match on, and the third spells it `sessionId`, which `\bsession\b` misses. So the detector has never recognised a missing session on any CLI we ship — which is exactly the hole you hit, and why widening the patterns (my first suggestion) would not have helped either.

Two things I would change in the approach:

**Auth still needs to propagate.** `-32603` is also what a broken token looks like, and with commit 3 in place a swallowed auth error drops the session *and* withholds the history — the user keeps typing into a blank session with no signal. There is a better discriminator sitting unused: `session/list` is on the same ACP surface, the agent has already answered on this connection, and asking only costs a round trip on a load that already failed. Not listed → gone. Still listed → keep the binding and surface the error.

**The `!sessionInvalidated` guard does not survive a save.** `consumeSessionInvalidation()` clears the flag, `save()` calls it, and `TabManager.destroy()` saves every tab on quit. Drop a session, quit before typing, and the next launch sees a conversation that looks like it never had one — full transcript replayed. The flag has to live in `providerState`.

I put all of it in #104, against the three twins, verified on the real CLIs. Your commits 2 and 3 are in there essentially unchanged — the diagnosis was yours, I only added the part that makes it hold. Happy to close that one in favour of yours if you would rather carry it.

sandsaber:
#104 is merged, so I'm closing this one — not because the diagnosis was wrong, but because it was right and now lives on main.

What carried over, essentially as you wrote it: history bootstrap reserved for a cold resume, and the `!sessionInvalidated` guard that makes the warmup path behave. What changed: the invalidation is persisted in `providerState.sessionDropped`, because the in-memory flag is consumed by the first `save()` and your guard could not survive a quit; and the `loadSession` catch asks `session/list` instead of widening the error class, so an auth failure still reaches the user with the binding intact.

The measurement in your description is what made this findable at all — thank you for chasing it down to a number.

greenkms:

PR #101, opened 2026-08-27
fix(chat): queue follow-ups as separate turns instead of merging them into one

greenkms:
## Problem

A message sent while a turn was still running did not join a queue. It was
merged into the one waiting slot: `mergeText` joined the two with a blank line,
and `mergeQueuedChatTurns` resolved `currentNotePath`, `vaultSearchContext` and
`projectWorkspaceContext` as `incoming ?? existing`. So a follow-up about one
note and a follow-up about another reached the model as a single turn carrying
only the second note. The boundary was lost, and so was the context the first
one was written against.

Two further consequences of the same single-slot design:

- A failed turn still drained the queue. `catch` appended `**Error:**` and
  `finally` drained anyway, so a dead CLI or an expired login spent the whole
  queue against nothing.
- Cancelling emptied the queue into the composer, and a steer that never landed
  was merged into whatever stood behind it.

Fixes #

## Root Cause

`ChatState` held `queuedMessage: QueuedMessage | null` - one slot, so a second
follow-up had nowhere to go but into the first. This is core behaviour in
`src/core/runtime/QueuedTurn.ts` and `src/features/chat/`, not a provider
quirk; it was reported on agy but reproduces everywhere.

The merge is not wrong in itself - it is wrong as the only option. Merging is
correct where the destination is genuinely one field, and the single-slot
design made every hand-off look like that case.

## Approach

**A real queue.** `ChatState.queuedMessage` becomes `ChatState.queue`, a FIFO of
`QueuedMessage` in a new `MessageQueue` (no DOM, no async, no provider
knowledge). Each entry keeps the `ChatTurnRequest` snapshot it was captured
with, which is what makes two follow-ups two turns. The drain takes the head
and only the head. The guards inside the existing `setTimeout(0)` are
untouched - with a queue they fire more often, not less.

**A failed turn no longer feeds the queue.** The queue is held on a failed turn
and on an explicit stop, with the same rule for both. The hold clears on resume
or once the queue empties, and a held queue cannot be armed when empty. `Stop`
no longer empties the queue into the composer.

**Discard and leaving a conversation are no longer one operation.** They were
the same method and could not stay that way: discard has to drop the text the
user asked to drop, while leaving a conversation must not silently destroy
hand-typed input. Discard drops; leaving hands the leftovers to the composer.
That hand-back is the one caller `mergeQueuedChatTurns` has left - the composer
is a single field, so merging is right there and only there. The function and
its tests are unchanged.

**A steer that never landed returns to the head**, with its snapshot intact,
instead of being merged into whatever stood behind it - the same defect in a
second place. This also covers a cancel arriving while a steer is in flight.

**The indicator became a list.** A row per entry with its own edit and discard,
so a typo in the third follow-up is fixable without losing the other two. One
waiting message keeps exactly the row it had: header, numbering and clear-all
appear only from the second message on, or once the queue is held. Steer sits
on the first row alone, because it takes the head and only the head. Rendering
moved into `QueueIndicator`; `InputController` was already the largest file in
the feature and the DOM it carried had nothing to do with deciding when to
drain.

**Editing a row keeps its place.** Pulling the third follow-up back into the
composer and sending it again returns it to third, not to the end: the row's
index is held while it is being edited and the queue re-inserts it there. The
held index is dropped if that same row is discarded meanwhile, shifts down if
an earlier one is, and is released if the turn ends first and the message
leaves as its own turn - a slot that no longer means anything must not catch
the next follow-up. It is never taken for a row that was not there.

**Presentation.** The list is separated from the composer by a hairline in the
composer's own colours - the neutral frame line at rest, the accent line once
the composer has focus. A queue sitting directly on top of the input needs the
boundary, and a boundary that stays grey while the frame lights up reads as a
rendering artefact rather than a divider.

`chat.queue.paused` is recorded with its reason, so a hold is visible in the
logs rather than inferred from a queue that stopped moving.

## Architecture

Architecture notes:

Nothing under `src/providers/` is touched, and no provider or ACP contract
changes. The queue is app-level chat behaviour and lives in
`src/features/chat/`: `queue/MessageQueue.ts` (state), `ui/QueueIndicator.ts`
(rendering), `controllers/InputController.ts` (when to drain). `MessageQueue`
is deliberately provider-neutral - it knows what is waiting and whether
draining is held, and nothing else.

`mergeQueuedChatTurns` in `src/core/runtime/QueuedTurn.ts` is kept and left
unchanged, with exactly one caller: handing the leftovers back to the composer.

The queue stays in memory. It does not reach
`.grimoire/grimoire-settings.json` and nothing new is written to disk.

## Security And Privacy

Security notes:

No security boundary change. No new persistence: queued follow-ups and their
turn snapshots live only in memory for the lifetime of the tab.

One behaviour is worth naming as a privacy improvement. Each entry now carries
the context snapshot it was captured with, so a follow-up written against one
note is no longer sent with a different note's contents attached, which is what
the `incoming ?? existing` merge produced. Holding the queue on a failed turn
also stops a queue of follow-ups being fired at a session whose state is
unknown.

The new debug event `chat.queue.paused` records the provider id, the queue size
and the reason. It does not record message text.

## Verification

Automated commands run:

```text
npm run typecheck
npm run lint
npm run build:release
npx jest --testPathPatterns "(MessageQueue|QueueIndicator|InputController|ChatState|ConversationController)"
```

to load and to open the view. The five suites above: 354 passed, 0 failed.

Full unit suite on Windows, this branch: 394 suites passed, 22 failed; 7442
tests passed, 36 failed, 8 skipped. The `main` baseline at `0d57617` on the
same machine: 392 suites passed, **the same 22** failed; 7419 passed, **the
same 36** failed, 8 skipped. The failures are the pre-existing path-separator
assertions and are unrelated to this change; the deltas are the two new suites
and the 23 new tests.

New coverage: `MessageQueue.test.ts` (ordering, removal by index, return to
head, pause and resume, and that two entries keep their own `currentNotePath`);
`QueueIndicator.test.ts` (single row versus list, header appearance, steer on
the first row only, removal by the row pointed at, held state); and two in
`InputController.test.ts` for the edited slot - released once its message is
sent as its own turn, never taken for a row that does not exist.

Six existing tests asserted the old merging behaviour and now assert the new
one. Listing them so they are not mistaken for breakage:

- `should append new message to existing queued message` -> `should queue a
  second message as its own entry instead of appending`
- `should merge images when appending to queue` -> `should keep images with the
  message they were attached to`
- `should clear queue on cancel` -> `should keep the queue and pause it when the
  user cancels`
- `should restore queued message to input when cancelling` -> `should leave the
  queue alone instead of dumping it into the composer`
- `should restore queued images to image context manager when cancelling` ->
  `should leave queued images on their message when cancelling`
- `should restore a pending steer message to input on cancel` -> `should return
  a pending steer message to the head of the queue on cancel`

The remaining references to the old field are mechanical, and
`QueuedTurn.test.ts` is untouched.

Manual verification:

Windows 11, Obsidian, MiMoCode provider. Two follow-ups typed during a running
turn arrive as two separate turns, each with the note it was written against;
the list renders one row per entry with per-row edit and discard; editing the
second row and re-sending returns it to second; the separator between the list
and the composer follows the composer border on focus in both light and dark
themes.

New i18n keys are in all ten locales.

## Generated Artifacts

No dependency change; `package.json` and `package-lock.json` are untouched.

## Review Checklist

- [ ] Protocol tests use real structured payloads and error codes.

The protocol-tests box is left unchecked because there are no protocol tests to
have: nothing under `src/providers/` is touched and no ACP payload is
constructed or parsed anywhere in this change.


sandsaber:
Reviewed this closely — the design is right, and the two consequences you found beyond the merge (a failed turn draining the queue, cancel emptying it into the composer) are the kind of thing that only shows up if you actually follow the state through. I put the branch plus six fixes in #106.

Four are gaps in the hold:

- `pause()` no-ops on an empty queue, and the failed-turn path pauses before `restorePendingSteerMessageToQueue()` — so a steered follow-up comes back to an unpaused queue and fires at the session that just died. `cancelStreaming` has the order right; that path does not.
- `pauseQueue()` never re-renders, and nothing else does on the failed path, so the hold has no header and no Resume — and Resume is the only way out of it.
- `processQueuedMessage` dequeues before the deferred `isStreaming` guard. Resume is clickable during the whole cancel round trip, so clicking it there takes the head out and then aborts — the message is gone.
- The slot held for an edited row is released by any send, including the one a drained message issues, so `[A,B,C]` edited at row 1 comes back as `[C,B]`.

One is a hand-off: `clearQueuedMessage` is wired to `drainQueueToComposer`, and `createNewConversation` calls it right after clearing the composer and the images — so the old conversation's follow-ups and their images land in the new one. I made conversation changes discard the queue, same as the composer text already is.

And one cosmetic: "Steering" ends up on row 0, which is the message *behind* the steer once it has left the queue.

Feel free to take any of it back into this branch if you would rather carry it — otherwise #106 supersedes.

sandsaber:
#106 is merged, so I'm closing this — the feature shipped, your branch carried it, and the seven commits are in main under your authorship.

For the record of what changed on the way in: four fixes to the hold (the steer returning before the pause instead of after, `pauseQueue` rendering so Resume is reachable, `processQueuedMessage` peeking rather than dequeuing before its guard, and the edit slot only being released by the send that actually held it), one to the hand-off (conversation changes discard the queue instead of pouring it into the next conversation), and the Steering label moving off the row behind the steer.

The part worth keeping: the single-slot design was the real bug, and the two consequences you traced from it — a failed turn draining the queue, cancel emptying it into the composer — are what made the shape of the fix obvious. Thanks for that.

PR #102, opened 2026-08-27
fix(providers): stop replaying the transcript into a session that replaced a lost one

greenkms:
## Problem

The MiMoCode PR fixed a resume failure that silently re-sent an entire
conversation into the session that replaced the lost one - measured at roughly
8.4M plan-usage credits on the first message after an Obsidian restart, against
~71k in the same conversation once it was running.

That defect is not MiMoCode-specific. `KimicodeChatRuntime`,
`OpencodeChatRuntime`, `GeminiChatRuntime`, `GrokChatRuntime` and
`QwenChatRuntime` compute `shouldBootstrapHistory` the same way, so the same
restart spends the same transcript on five more providers.

Fixes #100 

## Root Cause

Identical in all five, at `query()`:

```ts
const expectedSessionId = this.sessionId;
let shouldBootstrapHistory = previousMessages.length > 0
  && (!expectedSessionId || this.sessionInvalidated);
...
const cwd = getVaultPath(this.plugin.app) ?? process.cwd();
if (expectedSessionId && !this.sessionId) {
  shouldBootstrapHistory = previousMessages.length > 0;   // unconditional
}
```

Both branches lead to a replay, and the second one ignores why the session
went away.

The half that makes an obvious-looking fix insufficient: the session is
usually gone before `query()` starts. Tab warmup reaches `ensureReady()` on the
**bound** runtime - through `KimicodeRuntimeCommandLoader` and
`OpencodeRuntimeCommandLoader`, which are line-for-line analogues of the
MiMoCode loader, and through the workspace services and command loader for the
other three - so a `session/load` failure clears the binding while the composer
is still empty. `query()` then starts with no session id, which at that point is
indistinguishable from a cold resume that never had one. Reserving bootstrap for
`!expectedSessionId` alone therefore still replays on the path an editor restart
takes.

All five already set `sessionInvalidated = true` next to the
`clearActiveSession()` that follows a failed resume, so the state needed to tell
the two apart is present in every runtime. It was simply being read with the
opposite sign - as a *reason* to replay.

## Approach

One change per runtime, matching the MiMoCode fix:

```ts
let shouldBootstrapHistory = previousMessages.length > 0
  && !expectedSessionId
  && !this.sessionInvalidated;
```

and the unconditional re-set after `getVaultPath` is removed.

A session dropped by a load failure or a forced restart starts clean, and
recovering the earlier context is the user's to do. A conversation restored from
disk that never held a session still gets its history.

**Deliberately not ported.** The MiMoCode PR also widened `loadSession()`'s
catch from `isAcpMissingSessionError` to any `JsonRpcErrorResponse`, clearing
the binding instead of rethrowing. Kimicode and Opencode carry an explicit test
for the opposite behaviour - `preserves the saved session binding when
session/load fails transiently` - so that is a deliberate decision by whoever
wrote it, not an oversight, and reversing it in five runtimes I cannot exercise
live is not something to slip into a history-replay fix. It is also not needed
for this one: with the guard in place a stale binding no longer produces a
silent replay, it produces a visible error. Worth deciding separately, and
raised in the linked issue.

## Architecture

Architecture notes:

Five runtime files and their five test files; nothing shared is touched and no
signature changes. The duplication is pre-existing - these runtimes are
near-identical by construction - and this PR follows it rather than starting a
refactor into a shared helper in the same change. That extraction is worth
doing and worth doing on its own.

Sibling check is what produced this PR: MiMoCode was fixed first, the pattern
was traced across the provider tree, and these five are the remainder. Claude,
Codex and Antigravity do not use this session-resume shape.

## Security And Privacy

Security notes:

No security boundary change, and the same privacy improvement as the MiMoCode
fix, now on five more providers: a resume failure no longer re-sends the full
conversation - which in Obsidian routinely includes note text pulled in as
context - to a newly created remote session with no prompt and no indication.

Added tests assert on locally constructed strings only.

## Verification

Automated commands run:

```text
npm run typecheck
npm run lint
npm run build:release
npx jest --testPathPatterns "(KimicodeChatRuntime|OpencodeChatRuntime|GeminiChatRuntime|GrokChatRuntime|QwenChatRuntime)"
```

to load and to open the view.

The five runtime suites, this branch: 191 passed, 14 failed. The `main`
baseline at `0d57617` on the same machine: 181 passed, **the same 14** failed.
The failures are the pre-existing Windows path-separator assertions; the delta
is exactly the ten added tests.

Two tests per runtime: the session dropped before the turn must not replay, and
the cold resume that never held a session must still bootstrap. The second
direction is there on purpose - a guard like this is much easier to over-tighten
than to under-tighten, and without it the regression would be silent.

Manual verification:

Not run. None of these five providers is installed here; the behaviour was
reproduced and measured on MiMoCode only, and this PR is the mechanical port of
that fix to runtimes that share the code shape. Reviewers with any of the five
configured can check it in one pass: open a conversation with history, restart
Obsidian so `session/load` fails, and confirm the first message afterwards no
longer carries the transcript.

## Generated Artifacts

No dependency change.

## Review Checklist

- [ ] Protocol tests use real structured payloads and error codes.

No user-facing copy changes; no new i18n keys.

The protocol-tests box is left unchecked because the added tests do not
construct ACP payloads: the decision under test is taken before any request is
built, so they drive `query()` and assert on the prompt blocks it produces. The
existing protocol tests in these suites are untouched.

Issue #99

### DIFF
--- a/src/providers/gemini/runtime/GeminiChatRuntime.ts
+++ b/src/providers/gemini/runtime/GeminiChatRuntime.ts
@@ -259,8 +259,16 @@ export class GeminiChatRuntime implements ChatRuntime {
   ): AsyncGenerator<StreamChunk> {
     const previousMessages = conversationHistory ?? [];
     const expectedSessionId = this.sessionId;
+    // A session that was dropped must not replay the whole transcript into its
+    // replacement: that silently spends the conversation the user already paid
+    // for, without asking. `sessionInvalidated` is the difference between "we
+    // had a session and lost it" and "there was never one", and it is the only
+    // form of that answer which survives a drop that happened before query()
+    // ran - ensureReady() is also reached from warmup, so a null session id by
+    // itself does not mean a cold resume.
     let shouldBootstrapHistory = previousMessages.length > 0
-      && (!expectedSessionId || this.sessionInvalidated);
+      && !expectedSessionId
+      && !this.sessionInvalidated;
 
     if (!(await this.ensureReady())) {
       yield { type: 'error', content: t('chat.ui.errors.provider.startFailed', { provider: ProviderRegistry.getProviderDisplayNameOrId('gemini') }) };
@@ -275,9 +283,9 @@ export class GeminiChatRuntime implements ChatRuntime {
     }
 
     const cwd = getVaultPath(this.plugin.app) ?? process.cwd();
-    if (expectedSessionId && !this.sessionId) {
-      shouldBootstrapHistory = previousMessages.length > 0;
-    }
+    // Intentionally no history bootstrap here: if readiness dropped the session
+    // (session/load failure, forced restart), the new session starts clean and
+    // recovering the earlier context is the user's to do.
 
     if (!this.sessionId) {
       const sessionId = await this.createSession(cwd);
--- a/src/providers/grok/runtime/GrokChatRuntime.ts
+++ b/src/providers/grok/runtime/GrokChatRuntime.ts
@@ -518,8 +518,16 @@ export class GrokChatRuntime implements ChatRuntime {
   ): AsyncGenerator<StreamChunk> {
     const previousMessages = conversationHistory ?? [];
     const expectedSessionId = this.sessionId;
+    // A session that was dropped must not replay the whole transcript into its
+    // replacement: that silently spends the conversation the user already paid
+    // for, without asking. `sessionInvalidated` is the difference between "we
+    // had a session and lost it" and "there was never one", and it is the only
+    // form of that answer which survives a drop that happened before query()
+    // ran - ensureReady() is also reached from warmup, so a null session id by
+    // itself does not mean a cold resume.
     let shouldBootstrapHistory = previousMessages.length > 0
-      && (!expectedSessionId || this.sessionInvalidated);
+      && !expectedSessionId
+      && !this.sessionInvalidated;
 
     if (!(await this.ensureReady())) {
       logGrokDebug(this.plugin, 'query.ensureReady.failed', {
@@ -538,9 +546,9 @@ export class GrokChatRuntime implements ChatRuntime {
     }
 
     const cwd = getVaultPath(this.plugin.app) ?? process.cwd();
-    if (expectedSessionId && !this.sessionId) {
-      shouldBootstrapHistory = previousMessages.length > 0;
-    }
+    // Intentionally no history bootstrap here: if readiness dropped the session
+    // (session/load failure, forced restart), the new session starts clean and
+    // recovering the earlier context is the user's to do.
 
     if (!this.sessionId) {
       const sessionId = await this.createSession(cwd);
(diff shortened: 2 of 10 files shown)


sandsaber:
Same note as on #100, since this is the same guard across five runtimes.

`shouldBootstrapHistory = … && !this.sessionInvalidated` depends on a flag that does not survive a save: `consumeSessionInvalidation()` clears it, `save()` calls it, and `TabManager.destroy()` saves every tab on quit. So a session dropped during warmup, followed by a quit before the user types anything, comes back on the next launch looking like a conversation that never had a session — and the full transcript is replayed after all. The answer has to be persisted in `providerState` for the guard to hold.

I did that in #104, along with fixing the detection it sits on: `isAcpMissingSessionError` turns out to recognise a missing session on none of the CLIs we ship, because they report it as a bare `-32603 Internal error`. #104 covers the OpenCode/MiMoCode/Kimi Code trio only — Gemini, Qwen and Grok resume through their own paths and I could not put their CLIs on the wire, so this PR still has the ground it was standing on for those three.

sandsaber:
#104 landed, which covers OpenCode, MiMoCode and Kimi Code. This PR still has the other three to give — here is what it needs to rebase cleanly.

Dropping the `opencode` and `kimicode` hunks (and their tests) is all it takes; I checked, and `gemini`, `qwen` and `grok` merge onto main without a conflict.

One thing to know for those three: the `!this.sessionInvalidated` guard still leaks there. `consumeSessionInvalidation()` clears the flag, `save()` calls it, and `TabManager.destroy()` saves every tab on quit — so a session dropped during warmup, followed by a quit before the user types, comes back next launch looking like a conversation that never had one, and the transcript is replayed anyway. #104 fixed that for the trio with a `providerState.sessionDropped` marker; the same shape would port over, but their provider state and resume paths differ enough that I did not want to write it blind, and their CLIs are the ones I could not get on the wire.

So: correct as far as it goes, and worth merging for those three — just not the whole fix for them yet.

sandsaber:
#105 is merged, which finishes this for Gemini, Qwen and Grok — so I'm closing this one. Same as with #100: not because it was wrong, but because it was right and the rest of it is now on main.

Your `shouldBootstrapHistory` change carried over as written, for all three. What it needed underneath: the invalidation is persisted in `providerState.sessionDropped`, because `consumeSessionInvalidation()` is emptied by the first `save()` and `TabManager.destroy()` saves every tab on quit — so a drop followed by a quit came back next launch looking like a conversation that never had a session, and replayed anyway.

The part that surprised me while checking: Gemini and Qwen had a bare `catch { return false }`, so they were already treating an expired token as a lost session. Combined with your guard, that would have dropped the context silently. They now go through `isAcpSessionGone`, and it turns out both of them do name the reason in the error, so it costs nothing. Grok kept its old catch — its CLI would not answer a probe, and narrowing that blind seemed worse than leaving it.

Thanks for finding the class and for measuring it. Four PRs came out of that one report.

greenkms:

PR #105, opened 2026-08-28
fix(providers): finish the session-resume fix for Gemini, Qwen and Grok

sandsaber:
Supersedes #102. #104 covered the OpenCode/MiMoCode/Kimi Code trio; these three resume through their own code and were left with the same defect in a slightly different shape.

## What is different about these three

They never used the shared detector at all. Gemini and Qwen have a bare `catch { return false }` and the caller clears the session — so an expired token and a session that genuinely no longer exists are indistinguishable. Both silently drop the binding, and with #102's replay guard on top, both then drop the conversation's context with nothing said. That is the same trade #100 was heading for, already shipped here.

## Three fixes

**1. Discriminate.** Both now go through `isAcpSessionGone`. The widened patterns from #104 answer both without a round trip — I resumed an id neither has ever had:

| CLI | `session/load` on a missing session | Recognised? |
|---|---|---|
| Gemini CLI 0.53.1 | `-32603 "Internal error"`, `data.details: "No previous sessions found for this project."` | yes |
| Qwen Code 0.21.8 | `-32002 "Resource not found: session:<id>"` | yes |

`Authentication failed` is not recognised, and Gemini exposes no session listing to fall back on, so it propagates with the binding intact — which is the point. Qwen answers `session/list` too, so it has the listing as a backstop.

**2. Make the drop durable.** `consumeSessionInvalidation()` clears the flag, `save()` calls it, `TabManager.destroy()` saves every tab on quit. `providerState.sessionDropped` records it and is read back on load. Gemini and Qwen wiped `providerState` outright on invalidation, so the marker replaces that wipe; Grok's state filter had to learn the key.

**3. Stop the replay.** #102's fix, which only holds because of 2.

## Grok gets 2 and 3 only

Its CLI would not answer an ACP handshake off a bare probe, so what it reports for a session it no longer has is unobserved. Narrowing its catch blind would turn a silent recovery into a user-visible error on every stale resume — worse than the bug, and untestable. Its `AGENTS.md` records the reason and what would lift it.

## Verification

Live against the installed CLIs for Gemini and Qwen; their real payloads are checked in as fixtures. Each fix reverted in turn to confirm its tests fail without it — 2 for the detection, 6 for the durable marker, 3 for the replay guard.

Unit 7340 ✓ · integration 216 ✓ · typecheck ✓ · lint ✓ · `build:release` ✓.

https://claude.ai/code/session_01TDYHm4Ggxppte2w94Bdtak

### DIFF
--- a/src/providers/gemini/AGENTS.md
+++ b/src/providers/gemini/AGENTS.md
@@ -28,3 +28,8 @@ gemini --acp
 ```
 
 Custom CLI paths are stored per host under `providerConfigs.gemini.cliPathsByHost`. If no custom path exists, Grimoire launches `gemini` from PATH.
+
+## Session resume
+
+- On `session/load` failure, decide with `isAcpSessionGone` rather than treating any failure as proof the session is gone. Gemini CLI reports a missing session as `-32603 Internal error` with `data.details` naming the reason, which the shared message patterns recognise; it exposes no session listing, so anything unrecognised - transport, authentication, configuration - propagates with the binding intact.
+- A dropped session is recorded in `providerState.sessionDropped` and read back on load, because the in-memory flag is consumed by the first save. Never replay the transcript into a replacement session: history bootstrap is for a cold resume that never held a session id.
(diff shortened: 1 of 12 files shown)


greenkms:
Thanks for further analyzing my PR and getting to the root cause.

PR #106, opened 2026-08-28
fix(chat): queue follow-ups as separate turns, with the hold actually holding

sandsaber:
This is #101 plus six fixes found reviewing it. The feature is right — a queue instead of one merged slot, held when a turn dies — so the whole branch is here rather than a follow-up; #101 can close in favour of it.

## What the review turned up

**The steer escaped the hold.** `MessageQueue.pause()` deliberately no-ops on an empty queue, and the failed-turn path paused *before* returning the in-flight steer. One follow-up, steered, turn dies: the pause does nothing, the steer returns to an unpaused queue, and fires straight at the session that just failed — the exact thing the hold exists to prevent. `cancelStreaming` already had the right order; the failed path now matches it.

**The hold was invisible.** `pauseQueue()` never re-rendered, and nothing else on that path did either. The queue just stopped: no header, no **Resume** — and Resume is the only way out. The render moved into `pauseQueue`, where no caller can forget it.

**Resume during a cancel destroyed the head.** `cancelStreaming` renders Resume while `isStreaming` is still true, so it is live for the whole cancel round trip. `processQueuedMessage` dequeued first and checked `isStreaming` afterwards, inside the deferral — the message was already gone when the guard aborted. It now peeks, and removes the head only past the guard and only if it is still the same message.

**Leaving a conversation carried the queue into the next one.** `clearQueuedMessage` was pointed at `drainQueueToComposer`, which does the opposite of clearing: `createNewConversation` clears the composer and the images, then this refilled both from the conversation being left — old follow-ups merged in, their images re-attached after `clearImages()`, focus stolen. The composer's own text is already discarded on a switch; the queue belongs to the same conversation and now goes the same way.

**An edited row lost its place.** The slot held for a withdrawn row was released by *any* send, including the one a drained message issues — `[A,B,C]` edited at row 1 came back as `[C,B]`. Two shifts were missing besides: a drained head moves every later slot down, and withdrawing a second row leaves the merged text belonging at the earlier of the two positions.

**"Steering" labelled the wrong message.** A steer in flight has already left the queue, so row 0 is the message behind it — wearing the disabled Steering button. The steer is now announced above the list, where it is true.

## Verification

Each fix reverted in turn to confirm its tests fail without it. Unit 7370 ✓ · integration 216 ✓ · typecheck ✓ · lint ✓ · `build:release` ✓ (including `review:css`).

Two of the six were regressions this branch introduced rather than pre-existing — the steer ordering and the conversation hand-off — which is why they are worth landing together with the feature rather than after it.

https://claude.ai/code/session_01TDYHm4Ggxppte2w94Bdtak


greenkms:

PR #109, opened 2026-08-31
fix(claude): route auto-mode questions through user dialogs

jackycjrt:
## Motivation

I encountered this while migrating from Claudian to Grimoire. In Claudian's YOLO mode, Claude can still use `AskUserQuestion` when a task requires an explicit user decision. Grimoire already has a shared structured-question UI, but newer Claude Code versions may route these requests through `request_user_dialog` while bypass/auto mode is active, so they no longer reliably reach the existing `canUseTool` path.

Auto-approve should control whether tool actions require permission. It should not prevent the agent from asking the user to resolve a material ambiguity.

## Changes

- Register an `onUserDialog` handler for persistent and cold-start Claude queries.
- Advertise support for `permission_ask_user_question` dialogs.
- Route the dialog through Grimoire's existing structured-question callback.
- Preserve the SDK permission result while attaching the user's answers.
- Cancel unsupported or dismissed dialogs cleanly.

## Verification

- Claude runtime tests: 276 passed.
- Typecheck passed.
- Lint passed.
- Release build and isolated bundle verification passed.

### DIFF
--- a/src/providers/claude/runtime/ClaudeApprovalHandler.ts
+++ b/src/providers/claude/runtime/ClaudeApprovalHandler.ts
@@ -22,6 +22,7 @@ import type {
 } from '../../../core/types';
 import type { PermissionMode } from '../../../core/types/settings';
 import { buildPermissionUpdates } from '../security/ClaudePermissionUpdates';
+import { prepareAskUserQuestionInput } from './claudeAskUserQuestion';
 
 export interface ClaudeApprovalHandlerDeps {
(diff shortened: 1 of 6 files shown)


sandsaber:
Thank you for the PR, @jackycjrt — and for taking the time to write up the motivation so clearly. I pushed a follow-up commit to the branch with a few adjustments on top of your work, and I'll take it from here.

Much appreciated!

PR #110, opened 2026-08-31
fix(codex): enable user input in default mode

jackycjrt:
## Motivation

I encountered this while migrating from Claudian to Grimoire. In Claudian's YOLO mode, the agent can still ask a structured question when a task requires an explicit user decision. Grimoire already handles Codex `item/tool/requestUserInput` requests in its shared question UI, but Codex does not expose `request_user_input` in Default mode unless its official feature flag is enabled.

Auto-approve should control whether tool actions require permission. It should not prevent the agent from asking the user to resolve a material ambiguity.

Before this change, Codex could report that `request_user_input` was only available in Plan mode, or proceed by making an assumption instead of asking.

## Changes

- Start Codex app-server with the official `default_mode_request_user_input` feature enabled.
- Apply the flag to native Windows/macOS/Linux and WSL launch specifications.
- Keep the existing request routing and shared question UI unchanged.

## Compatibility

The flag is experimental and remains owned by Codex. It was verified with Codex CLI 0.117.0 and 0.147.0. Codex currently treats Default-mode questions as non-blocking and may resolve an unanswered request after its provider-defined timeout.

## Verification

- Codex launch and request-router tests: 32 passed.
- Typecheck passed.
- Lint passed.
- Release build and isolated bundle verification passed.
- Verified a live Grimoire app-server process starts with the feature enabled.

### DIFF
--- a/src/providers/codex/runtime/CodexLaunchSpecBuilder.ts
+++ b/src/providers/codex/runtime/CodexLaunchSpecBuilder.ts
@@ -14,12 +14,17 @@ export interface BuildCodexLaunchSpecOptions {
   resolveDefaultWslDistro?: () => string | undefined;
 }
 
-// Keep Grimoire's structured AskUserQuestion UI available while Codex is in
-// its normal/default collaboration mode (the toolbar's Safe and Auto modes).
+// Codex keeps request_user_input behind an experimental feature flag in its
+// default collaboration mode (the toolbar's Safe and Auto modes), so without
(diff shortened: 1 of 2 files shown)


sandsaber:
Thank you for this one as well, @jackycjrt — reproducing it against a live app-server made the problem very clear, and the pointer to the official Codex feature flag saved a lot of digging. I pushed a follow-up commit to the branch and will take it from here.

Thanks again for both contributions!

PR #115, opened 2026-09-02
feat(antigravity): let the user attach images to an agy turn

greenkms:
## What

Pasting or dropping an image into an Antigravity chat is refused with "Image
attachments are not supported by this provider." This makes it work: the
attachment is written to a temp file and its absolute path is named in the
prompt.

`supportsImageAttachments` was `false` for Antigravity alone (every other
provider has it `true`), so `ImageContextManager` is disabled through
`tabProviderUI.ts` and `ImageContext.addImageFromFile()` bails out before the
runtime is ever asked. The flag was correct at the time: `agy` exposes no image
flag, and its stream-json `user` event carries `content` as a plain string
rather than content blocks, so an attachment has no channel to travel in.

## Why this transport

`agy` does have file access. Measured 2026-09-03 against agy on Windows 11
(CLI resolved from `%LOCALAPPDATA%\agy\bin\agy.exe`), using a generated PNG of
a red circle placed in `%LOCALAPPDATA%\Temp`, i.e. outside the vault:

| Run | Conditions | Result |
|---|---|---|
| a | `agy --print`, no `--add-dir`, cwd = vault | exit 0, "The image depicts a red circle on a white background." |
| b | `--input-format stream-json --output-format stream-json`, prompt on stdin, `--add-dir <vault>` | `status: SUCCESS`, response "The image contains a red circle on a white background.", 8.3 s |

So the agent opens an absolute path from the prompt even when the file lives
outside the workspace and no `--add-dir` is passed. That is what makes this
patch small: the `addDirPath` contract is untouched, and attachments do not
depend on the capability probe confirming `--add-dir` — which matters because
the Windows build prints `--help` to stderr only, and the probe fails closed
for the turn when it cannot conclude.

Two alternatives were dropped. A temp directory inside the vault would have
been needed only if agy refused paths outside the workspace; run (a) shows it
does not, and writing user files into the vault to work around a limit that
does not exist is not worth it. Real content blocks in the stream-json input
are not documented by agy, and the provider's own rule is to prefer live CLI
output over guessed schemas.

## Behavior

- `AntigravityImageAttachments.attachAntigravityImages()` writes each
  `image/*` attachment into a fresh `os.tmpdir()` directory as
  `<n>-<sanitized name>` and appends the absolute paths to the prompt. The
  positional prefix keeps two attachments sharing one name apart; the name is
  sanitized because a pasted name is user input and separators in it would
  escape the directory that cleanup deletes.
- The directory is removed in the turn's `finally` on every exit — success,
  failure and cancel alike. This is deliberately unlike `--log-file`, which is
  preserved on failure for diagnosis: these files hold user data and carry
  nothing diagnostic.
- An attachment that cannot be written is reported through `recordDebugLog`
  (`print.imageAttachmentFailed`) and skipped, so one bad file does not cost
  the user the answer to the rest of the turn.
- A turn without images creates no directory at all.

`toAttachmentFilename` is duplicated rather than imported from
`src/providers/codex/`: it is private there (exported for tests only), and the
provider boundary in `src/providers/antigravity/AGENTS.md` asks for
Antigravity behavior to stay inside its own directory.

## Tests

`tests/unit/providers/antigravity/AntigravityImageAttachments.test.ts` (new,
12 cases) covers decoded bytes reaching the file, the path landing in the
prompt, collision-free naming, skipping non-image media types, surviving a
failed write, cleanup, idempotent cleanup, and filename sanitization including
`../../etc/passwd.png`.

`AntigravityChatRuntime.test.ts` adds four end-to-end cases over the mocked
CLI: the path reaches agy's stdin user event and the file exists while the run
is open; the directory is gone after a successful turn; gone after a failed
turn; and no directory is created for a turn without images.

`registration.test.ts` pins `supportsImageAttachments: true`.

Locally: `npm run typecheck` and `npm run lint` clean, `npm run build:release`
clean. The full suite is green except 23 suites that already fail on this
Windows checkout before the change (POSIX path assumptions in grok, kimicode,
mimocode, opencode, qwen and utils tests) — identical set and identical 36
failing cases with and without this patch.

## Not in scope

Images in auxiliary flows (title generation, inline edit — Antigravity has no
auxiliary runner), images in agent output, and any change to `--add-dir`.

## Follow-up: attachment names outside ASCII

The first commit reused the Codex sanitizer, whose whitelist is
`[A-Za-z0-9._-]`. That turns `дефект-слой-1.png` into `_____-____-1.png`, and
the name is not decoration in this transport: agy is handed a path rather than
an attachment, so the basename is the only label the image carries.

Measured 2026-09-03 over the stream-json transport with a PNG named
`дефект-слой-1.png` in `%LOCALAPPDATA%\Temp`:

```
"status":"SUCCESS","response":"дефект-слой-1.png\nBlue rectangle\n"
```

agy opened the non-ASCII path and echoed the name verbatim, so the
transliteration was pure loss. The second commit replaces the whitelist with a
blocklist scoped to what actually breaks a write, applied identically on every
host so one attachment cannot produce two different names depending on the
machine:

- both path separators (`/` and `\`), because a separator or `..` in a pasted
  name would escape the directory cleanup deletes;
- characters Windows rejects (`<>:"|?*`), trailing dots and spaces it strips
  silently, and reserved device names (`CON.png` → `_CON.png`);
- control characters and zero-width / bidirectional overrides, dropped rather
  than replaced so they add no padding to a name the agent reads back;
- a byte budget: NAME_MAX counts bytes and 300 Cyrillic letters are 600 of
  them, so truncation walks code points and never cuts a character in half.

Everything else is preserved, including non-ASCII letters. Nine more unit
cases cover Cyrillic and CJK names, both separators, device names, overrides,
dot-only names and byte truncation.

## Verified by hand

On the installed build, in a real Antigravity chat (Obsidian restarted, agy on
Windows 11):

- a single PNG: the agent opened it and described the contents correctly;
- several attachments in one turn, with non-ASCII and awkward file names: all
  of them reached the agent, and the names survived intact;
- one of them an animated GIF (`image/gif` is a first-class type in
  `ImageContext`): handled normally, no special casing needed.


greenkms:
CI on this PR fails at Audit dependencies, before Lint, Typecheck, Test and Build ever run — so nothing here reflects the diff. This branch touches no dependency: git diff origin/main..HEAD -- package.json package-lock.json is empty.

npm audit reports two transitive advisories, both under @modelcontextprotocol/sdk: fast-uri@3.1.5 (high, GHSA-5jgf-p345-68v8 and three more, via ajv) and qs@6.15.3 (moderate, GHSA-x5fp-wj9c-mxmx, via express). The fast-uri advisory was published 2026-09-02; the last green CI on main ran 2026-08-31 with the same lockfile. Since npm audit queries the advisory database live, main fails the same way today.

I left the lockfile alone rather than folding a dependency bump into a feature PR. Happy to rebase once the bump lands. Windows-unit on this branch is green, and lint, typecheck, build:release and the full suite pass locally.

sandsaber:
Reviewed this and liked the approach — the transport is backed by real measurements rather than a guessed schema, and the tests actually check the file exists *while* agy is running, which is the part that matters.

I pushed three follow-up commits rather than leaving a list of nits:

- **A dropped attachment was invisible.** A failed write only went to `recordDebugLog`, and debug logging is off by default, so the user just saw agy answer about fewer images than they attached. The names now go out as a `notice` chunk too.
- **The bundle was built outside the turn's `try`.** A throw while sanitizing escaped `query()` as a raw TypeError instead of reaching the localized catch. Moved it inside.
- **The byte budget only covered the stem.** A name like `a.` + 300 characters spent the whole budget on its "extension", so the write failed with ENAMETOOLONG and the image was silently dropped. A tail longer than any real extension is now treated as part of the name. While in there, the positional prefix moved into the helper — that's where the budget can account for it, and it also showed the device-name escape was dead code (`1-CON.png` is already a file, not a device), so `CON.png` keeps its own name now instead of gaining an underscore.

Plus one thing this PR uncovered next door: `toAttachmentFilename` in the Codex runtime reads `attachment.filename`, but the field is called `name` — so Codex has been writing every attachment as `image-<n>` and throwing the user's file name away. Fixed in a separate commit.

The red `validate` check is not from this PR — it's the `npm audit` failure that's been on `main` since the September advisories. #117 fixes it; update this branch once that lands and the check should go green.

PR #119, opened 2026-09-03
fix(chat): keep image attachments viewable, and let Escape close the viewer

greenkms:
Two defects in image attachments, found while using them on a live vault.
They are independent; the commits are separate and either can be dropped.

## 1. The full-size viewer went blank once a turn ended

Reproduction: attach an image, click it - the full-size view opens. Send
the turn, wait for the answer, click the same image - the overlay opens
empty, with a broken-image placeholder.

`updateConversation` cleared `ImageAttachment.data` for every message on
every call, on the premise its comment states: "data is persisted by
SDK". That premise holds for exactly one provider. Claude's transcript
stores user attachments as base64 `image` blocks and `extractImages`
reads them back on hydration. No other provider does, and Antigravity
cannot: attachments reach `agy` as temp files that the turn deletes, so
the message object is the only copy of the bytes.

The timing in the reproduction follows from that. An `` already in
the document keeps working because its `src` holds the data URI, but the
viewer rebuilds `src` from `image.data` on every click.

Two consequences went past the viewer:

- the clearing fired far from anything that saves messages. Any field
  update reached it, including `{ titleGenerationStatus: 'pending' }`, so
  a method whose contract is "update conversation fields" silently
  emptied message payloads;
- a draft restored through `setImages()` - the path taken when a send
  fails - would resend empty files, with nothing in the UI to say so.

`releaseRestorableImageData` now gates the release on a history-service
capability. The default is to keep the bytes: a provider that forgets to
declare `restoresImageAttachmentData` wastes memory, while one that
wrongly claims it loses user data, so absence means keep. Only
`ClaudeConversationHistoryService` declares it.

## 2. Escape closed the viewer and cancelled the streaming turn

Reproduction: send a turn, open an attached image while it streams, press
Escape. The overlay closes and the turn is interrupted.

Escape means two things in the chat view, and the viewer's handler was
registered on the document in the bubble phase without `preventDefault`,
so the view scope acted first.

Capture phase plus `preventDefault` is the obvious fix and it does not
work - measured on a live vault, not reasoned about. Which listener runs
first is a property of registration order, and Obsidian registers its own
document listener when the app boots, before any plugin can.

So the cancel paths ask rather than race. An open viewer registers itself
in `imageViewerStack`; both Escape handlers that would cancel a turn -
the view scope and the tab's input handler - close the topmost viewer
first and stop there. The outcome is the same under either ordering: the
viewer's listener closes and consumes the key, or a cancel path closes
the viewer and declines to cancel. A stack rather than a single slot
keeps nesting honest.

## Test harness

The mock element's fake `ownerDocument` dropped the third argument of
`addEventListener`. Capture-phase registration was therefore invisible to
tests and indistinguishable from bubble - the assertion that would have
caught this class of bug could not be written. It now forwards `options`.

## Verification

- `npm run typecheck`, `npm run lint`, `npm run build:release`: exit 0.
- Full `node scripts/run-jest.js` on a Windows host: 38 failing tests
  before and after this branch, the same set of suites - these are
  pre-existing POSIX-path failures on this platform, measured on a clean
  `origin/main` rather than taken on trust.
- New unit tests cover the capability gate, the pending-fork exception,
  the viewer stack, and Escape consumption in both viewers.
- Both defects were re-checked by hand on the installed build: the viewer
  now opens the image before and after the answer, and Escape closes the
  overlay without touching the turn.

## Not addressed

Antigravity shows an attached image to the agent for the turn it was
attached to and no later: `agy` has no image blocks in its stream-json,
so a follow-up turn has nothing to resend. That is a property of the
transport, not of this change.

### DIFF
--- a/src/core/bootstrap/imageAttachmentLifetime.ts
+++ b/src/core/bootstrap/imageAttachmentLifetime.ts
@@ -1,0 +1,50 @@
+/**
+ * In-memory lifetime of `ImageAttachment.data`.
+ *
+ * Attachment bytes are the largest thing a conversation holds - a single
+ * screenshot is megabytes of base64, and every open conversation keeps its
+ * messages in memory - so they are released once the conversation is saved.
+ *
+ * That is only safe when the provider can put them back. A provider whose
+ * transcript stores the image itself rehydrates the bytes in
+ * `hydrateConversationHistory`; a provider that never sees an image as data
+ * cannot. Antigravity is the second kind: attachments reach `agy` as temp
+ * files that the turn deletes, so this object is the only copy, and clearing
+ * it loses the image for the rest of the session - the full-size viewer shows
+ * an empty frame, and a restored draft would resend empty files.
+ *
+ * The release is therefore opt-in per provider, and the default is to keep the
+ * bytes: a provider that forgets to declare the capability wastes memory,
+ * while one that wrongly claims it loses user data.
+ */
+
+import type { ProviderConversationHistoryService } from '../providers/types';
+import type { Conversation } from '../types';
+
+/**
+ * Clears attachment bytes the provider is able to restore, and leaves every
+ * other attachment intact.
+ */
+export function releaseRestorableImageData(
+  conversation: Conversation,
+  historyService: ProviderConversationHistoryService,
+): void {
+  if (historyService.restoresImageAttachmentData !== true) {
+    return;
+  }
+
+  // A pending fork's deep-cloned images have not reached provider storage yet,
+  // so nothing would restore them.
+  if (historyService.isPendingForkConversation(conversation)) {
+    return;
+  }
+
+  for (const message of conversation.messages) {
+    if (!message.images) {
+      continue;
+    }
+    for (const image of message.images) {
+      image.data = '';
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


sandsaber:
Closing in favour of #126, which takes a different route to the same two defects.

The empty viewer turned out not to need a provider capability at all. Moving attachment bytes into a content-addressed store under `.grimoire/attachments/` means nothing releases them, so no provider has to declare whether it could put them back — and the release this PR gated was also unsafe for the one provider that declared it: it ran while the conversation was open, and `hydrateConversationHistory` returns early for a conversation already in `hydratedConversationIds`, so on Claude the viewer stayed empty for the rest of the session even after switching away and back.

Your Escape work is carried over as its own commit, authorship intact — the diagnosis that racing Obsidian's document listener is the wrong shape of solution was right, and the viewer stack is what #126 uses. Three review points were folded in on top: the stack now pops before closing so a `close` that fails to unregister cannot swallow every later Escape; both viewers delegate to the stack and use `stopImmediatePropagation`, because they register on the same document and `stopPropagation` does not stop siblings on the same node, so one Escape was closing every open viewer oldest-first; and the view scope consults the stack only for an Escape nobody consumed.

The `mockElement` fix came along too — it also surfaced a real bug: neither Codex nor Antigravity checked that an attachment had bytes before writing it to a temp file, so a resend after the clearing in `updateConversation` handed the CLI a zero-byte file described as a screenshot. Same defect as the viewer, in a second place.

Thanks for the measurements and the write-up on both — they made the shape of the problem obvious.

PR #120, opened 2026-09-04
feat(chat): generate conversation titles in the UI locale

greenkms:
### Problem

Auto-generated conversation titles are always in English, regardless of the
plugin's `locale` setting. `TITLE_GENERATION_SYSTEM_PROMPT` is English and
nothing tells the model which language to answer in, so a non-English vault
gets English tab titles.

Two adjacent issues surfaced while fixing that:

- Weaker title models (e.g. a small local model routed through OpenCode) ignore
  *"return ONLY the raw title text"* and prepend a label — `Generated title:`,
  `Title -`, `Заголовок:` — or a short preamble / markdown heading, which then
  becomes the tab name.
- `ClaudeTitleGenerationService` carried byte-identical private copies of
  `buildTitleGenerationPrompt` and `parseTitleGenerationResponse`, so any
  hardening of the shared helpers silently skipped the Claude title path.

### Changes

**1. Locale-aware system prompt** (`2d00d56`)

New `buildTitleGenerationSystemPrompt(locale?)` in
`src/core/prompt/titleGeneration.ts` appends an explicit language directive
derived from the plugin locale (`getLocale()` + `SUPPORTED_LOCALES`,
e.g. `ru` → *"Write the title in Russian"*). The base rules stay in English —
models follow them more reliably — and only the final title text is translated.
`resolveTitleLanguageName()` is exported alongside it.

The directive is emitted for every locale, `en` included: without it a model
mirrors the conversation language, so an English UI would get Russian titles for
a Russian conversation. Making it explicit fixes both directions.

`TITLE_GENERATION_SYSTEM_PROMPT` is unchanged and still exported.

**2. Response noise stripping** (`0a420e1`)

`parseTitleGenerationResponse()` now strips leading label prefixes
(`Generated:` / `Generated title -` / `Title:` / localised `Заголовок:`,
`Название:`, `Тема:`), markdown headings and bullets, and `Here is the title:`
style preambles; then takes the first real line; and only then trims wrapping
quotes and trailing punctuation. A label is only stripped when followed by a
separator, so a legitimate title like `Title screen crash` is untouched.

**3. Drop the duplicated Claude helpers** (`c07be20`)

`ClaudeTitleGenerationService` now calls the shared `buildTitleGenerationPrompt`
and `parseTitleGenerationResponse` instead of its private `truncateText` +
inline template and `parseTitle`. The copies were behaviourally identical
(500-char input truncation, 50-char output truncation, same quote/punctuation
trimming), so this is a no-op for already-clean responses — the existing
`TitleGenerationService` tests for quotes, trailing punctuation, 50-char
truncation and long-message truncation still pass unchanged — and the Claude
path now gets the same response hardening as every other provider.

Both title services now build the system prompt through the shared helper:
`QueryBackedTitleGenerationService` (Codex, Grok, OpenCode, Gemini, …) and
`ClaudeTitleGenerationService`.

### Tests

`tests/unit/providers/claude/prompt/titleGeneration.test.ts` — 15 cases (was 4):
locale resolution, unknown-locale fallback, explicit-locale argument,
prefix/preamble stripping, label on its own line, quotes and punctuation,
empty → `null`.

Full suite, measured on this branch and on `origin/main` in the same working
copy:

| | Suites failed | Tests failed | Passed | Total |
|---|---|---|---|---|
| `origin/main` | 24 | 37 | 7598 | 7643 |
| this branch | 24 | 37 | 7609 | 7654 |

Identical failure counts — the 37 are pre-existing and unrelated; +11 new
passing tests. `tsc --noEmit` and `eslint --max-warnings=0` clean.

### Relationship to #118

#118 (`feat(chat): make the fallback conversation title readable`) and this PR
are complementary, not competing approaches to the same thing. They touch
disjoint files:

- **#118** improves the *fallback* title — the local, no-LLM heuristic in
  `src/core/prompt/fallbackTitle.ts` used by
  `ConversationController.generateFallbackTitle()` when generation is disabled
  or fails.
- **This PR** improves the *generated* title — the system prompt sent to the
  model and the parsing of its response, in `src/core/prompt/titleGeneration.ts`
  and the two title services.

They can merge in either order without conflict.

### Verification

Checked on a live vault with `locale: ru`: OpenCode routed to a local
`qwen2.5-coder` model produces correct Russian titles; a smaller model that
previously emitted a `Generated` prefix is now parsed cleanly.

### DIFF
--- a/src/core/auxiliary/QueryBackedTitleGenerationService.ts
+++ b/src/core/auxiliary/QueryBackedTitleGenerationService.ts
@@ -1,7 +1,7 @@
 import {
   buildTitleGenerationPrompt,
+  buildTitleGenerationSystemPrompt,
   parseTitleGenerationResponse,
-  TITLE_GENERATION_SYSTEM_PROMPT,
 } from '../prompt/titleGeneration';
 import type {
   TitleGenerationCallback,
@@ -45,7 +45,7 @@ export class QueryBackedTitleGenerationService implements TitleGenerationService
       const text = await runner.query({
         abortController,
         model: this.options.resolveModel?.(),
-        systemPrompt: TITLE_GENERATION_SYSTEM_PROMPT,
+        systemPrompt: buildTitleGenerationSystemPrompt(),
       }, buildTitleGenerationPrompt(userMessage));
       const title = parseTitleGenerationResponse(text);
       await this.safeCallback(
--- a/src/core/prompt/titleGeneration.ts
+++ b/src/core/prompt/titleGeneration.ts
@@ -1,4 +1,4 @@
-import { SUPPORTED_LOCALES } from '../../i18n/constants';
+import { getLocaleInfo } from '../../i18n/constants';
 import { getLocale } from '../../i18n/i18n';
 import type { Locale } from '../../i18n/types';
 
@@ -22,8 +22,7 @@ export const TITLE_GENERATION_SYSTEM_PROMPT = `You are a specialist in summarizi
  * Falls back to English for unknown locales.
  */
 export function resolveTitleLanguageName(locale: Locale = getLocale()): string {
-  return SUPPORTED_LOCALES.find((entry) => entry.code === locale)?.englishName
-    ?? 'English';
+  return getLocaleInfo(locale)?.englishName ?? 'English';
 }
 
 /**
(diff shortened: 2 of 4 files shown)


sandsaber:
Pushed review fixes to this branch — details in the commit message.

The blocking one: taking the first line is only safe once the announcement ahead of it is gone, and the prefix list could not recognise every phrasing. `"Here is your title:\nFix the parser"` parsed to `Here is your title` — the answer was lost, where before the stripping existed it at least survived inside the text.

- The preamble pattern accepts the determiners models actually write (`your`, `my`, `one`), so `Here is your title: X` is stripped inline.
- A line that only announces the title is skipped when another line follows. A response that is nothing but such a line still uses it, so `Debugging:` is unchanged, and a colon line long enough to be a title in its own right is not skipped.
- `MAX_NOISE_STRIP_PASSES` is derived from the prefixes that can stack rather than left as a bare `4`.
- `resolveTitleLanguageName` goes through the existing `getLocaleInfo` instead of a second `find` over the same table.

Left alone deliberately: `Sure! Fixing the parser` is not stripped. It was not among the phrasings this PR claimed, and widening the denylist on speculation risks eating real titles.

I did check the thing that worried me first — backtracking in `TITLE_NOISE_PREFIXES`. 60 KB of whitespace and 100 KB of prose both parse in 0 ms; the patterns are anchored and there is no problem there. 

Full gate on the branch merged with current `main`: `npm test`, `typecheck`, `lint` all clean.

PR #124, opened 2026-09-05
feat(chat): give every title source the same 100-character budget

### LINKED ISSUE #125: A generated conversation title stops at 50 characters while the tab allows 100, because every producer carries its own budget
## Summary

`MAX_TAB_TITLE_LENGTH` is 100, and a tab title typed by hand may use all of it. A title
*produced by Grimoire* never can: every producer applies its own 50-character budget
before the tab ever sees the string. Half of the field is unreachable by design, and
because the cut is a plain `slice`, it lands mid-word.

The 50 predates `MAX_TAB_TITLE_LENGTH = 100` and was copied rather than derived, so the
two drifted apart. Today a title's maximum length depends on which mechanism happened to
produce it — model, fallback heuristic, or fork.

## What happens now

Four independent constants decide how long a conversation title may be, and none of them
agrees with the 100 the tab allows (all references on `main` at `a5f9a98`):

| Producer | Budget | Where |
| --- | --- | --- |
| Model prompt and its parser | `MAX_TITLE_LENGTH = 50`, prompt says `max 50 chars`, cut is `slice(0, MAX_TITLE_LENGTH - 3) + '...'` | `src/core/prompt/titleGeneration.ts:2`, `:6`, `:39` |
| `ClaudeTitleGenerationService.parseTitle` | private copy: `if (title.length > 50) title.substring(0, 47) + '...'` | `src/providers/claude/auxiliary/ClaudeTitleGenerationService.ts:110` |
| Fallback heuristic | `DEFAULT_MAX_TITLE_LENGTH = 50` | `src/core/prompt/fallbackTitle.ts:3` |
| `buildForkTitle` | local `const MAX_TITLE_LENGTH = 50` | `src/features/chat/tabs/TabManager.ts:860` |
| The tab itself | `MAX_TAB_TITLE_LENGTH = 100` | `src/features/chat/tabs/types.ts:81` |

Two consequences follow from the duplication itself, independently of the value:

- the instruction sent to the model (`max 50 chars`) and the invariant enforced on its
  answer are separate literals, so they can drift again without any test noticing;
- the Claude path is truncated by its own private parser, so a fix applied to the shared
  prompt code does not reach it.

## Evidence

A title generated in normal use, stored and displayed in full — this is not a display
ellipsis, the saved string really ends here:

```
Диагностика ошибки 500 Internal server error Cl...
```

That is exactly 50 characters, which is what `slice(0, 47) + '...'` produces. The subject
of the conversation — "Claude Code" — is not shortened but gone, while the tab had 50
unused characters available.

## Directions

1. **One budget, derived everywhere.** A single shared constant with the truncation
   implementation next to it, consumed by the prompt text, the parsers, the fallback
   heuristic, the fork title and `MAX_TAB_TITLE_LENGTH`. The prompt limit should be
   interpolated from the same constant so the instruction and the enforcement cannot
   diverge again.
2. **Reuse the word-boundary cut from #118** rather than writing a third truncation. It
   already avoids splitting a surrogate pair and only prefers a word boundary while the
   result keeps enough of the budget, so text that opens with one long token (a path, a
   URL) does not collapse to a single word.
3. **Keep the parser as a safety net.** Weaker local models ignore length instructions
   often enough that truncation cannot be dropped in favour of the prompt alone.
4. **Decide the ceiling deliberately.** Raising it to 100 also removes the brevity
   pressure that `max 50 chars` applied, so titles may drift longer. If that reads badly
   in practice, the instruction can be `aim for about 50 characters; never exceed 100` —
   the enforcement stays at the shared budget either way.

I have a change ready for this and will link it here. #124

greenkms:
### Problem

A tab title may be 100 characters long (`MAX_TAB_TITLE_LENGTH`), but a title that
comes from a model can never exceed 50: the system prompt asks for "max 50 chars"
and the parser hard-truncates at 50 with `slice(0, 47) + '...'`. Half of the field
is unreachable by design, and the cut lands mid-word:

```
Диагностика ошибки 500 Internal server error Cl...
```

That is not a display ellipsis — the stored title really ends there, and "Claude
Code" is gone from it. The fallback heuristic has its own 50 (`fallbackTitle.ts`),
`buildForkTitle` a third one, and `ClaudeTitleGenerationService` a fourth copy inside
its private parser. Four independent constants decide how long a conversation title
may be, none of them agreeing with the 100 the tab itself allows.

### Root cause

The 50-character budget predates `MAX_TAB_TITLE_LENGTH = 100` and was duplicated
rather than derived, so the tab-title limit and the title-producing code drifted
apart. Because each producer truncates on its own, a title's length depends on which
mechanism happened to produce it — model, fallback heuristic, or fork.

### Approach

One budget, one truncation implementation, derived everywhere.

`src/core/prompt/titleLength.ts` (new) owns `MAX_TITLE_LENGTH = 100` and
`truncateTitleOnWordBoundary`. The truncation logic is **not** new code: it is the
word-boundary cut introduced with #118 in `fallbackTitle.ts`, moved out so the other
producers can use it instead of their own `slice`. It keeps that implementation's two
properties — a cut never leaves half a surrogate pair behind, and a word-boundary cut
is only preferred while it retains at least 30% of the budget, so text opening with
one very long token (a path, a URL) does not collapse to a single word.

Consumers now derive from it:

| Producer | Before | After |
|---|---|---|
| Generated title (`titleGeneration.ts`) | prompt "max 50 chars", `slice(0, 47) + '...'` | prompt "max 100 chars", shared word-boundary cut |
| `ClaudeTitleGenerationService.parseTitle` | private copy, hardcoded 50 | shared helper |
| Fallback heuristic (`fallbackTitle.ts`) | `DEFAULT_MAX_TITLE_LENGTH = 50` | derives from `MAX_TITLE_LENGTH` |
| `MAX_TAB_TITLE_LENGTH` | independent literal `100` | derives from `MAX_TITLE_LENGTH` |
| `TabManager.buildForkTitle` | local `const MAX_TITLE_LENGTH = 50` | `MAX_TAB_TITLE_LENGTH` |

The prompt limit is interpolated from the same constant, so the instruction sent to
the model and the invariant enforced on its answer cannot drift again. The parser
remains a safety net rather than the primary mechanism — weaker local models ignore
length instructions regularly enough that the truncation has to stay.

`ClaudeTitleGenerationService` is included because on `main` it still carries a private
parser copy with its own 50; without that hunk the fix would not reach the Claude path
at all. (#120 removes that duplication entirely — see below.)

### Architecture

The new module sits in `src/core/prompt/`, next to the title helpers it serves, and is
consumed by shared feature code and by one provider service that currently duplicates
shared logic. No provider-native behavior moves into shared code, and no provider
contract changes.

### Security And Privacy

No security boundary change. The change affects string length limits only: no
filesystem, process, credential, MCP, permission or network behavior is involved.
Model output is still treated as untrusted text and is truncated before use.

### Tests

- `tests/unit/core/prompt/titleLength.test.ts` (new) — the budget value, text that
  fits left untouched, word-boundary cut, hard-cut fallback when one word would eat
  the budget, surrogate-pair safety, a budget too small to hold an ellipsis, the
  prompt carrying the shared budget, and an over-long model answer cut on a word
  boundary.
- `tests/unit/core/prompt/fallbackTitle.test.ts` — cases that asserted the old default
  now pass `maxLength: 50` explicitly, so the word-boundary behaviour they were written
  for is still covered, plus new cases for the 100-character default.
- `tests/unit/features/chat/services/TitleGenerationService.test.ts`,
  `tests/unit/features/chat/controllers/ConversationController.test.ts`,
  `tests/unit/features/chat/tabs/TabManager.test.ts`,
  `tests/unit/providers/claude/prompt/titleGeneration.test.ts` — truncation
  expectations moved to the shared budget.

Full unit suite, measured in the same working copy:

| | Suites failed | Tests failed | Passed | Total |
|---|---|---|---|---|
| `origin/main` | 22 | 35 | 7415 | 7458 |
| this branch | 22 | 35 | 7426 | 7469 |

Identical failure counts; +11 new passing tests. The 35 pre-existing failures are
Windows path-separator expectations in provider tests and are unrelated.

### Relationship to the other open PRs

- **#120** (`generate conversation titles in the UI locale`) touches the same two files
  and **conflicts textually** with this PR: it replaces
  `ClaudeTitleGenerationService`'s private parser with the shared one, while this PR
  fixes that private parser in place. The resolution is trivial and I have already
  exercised it locally: take #120's version of the service (its shared parser already
  carries this PR's truncation) and keep both import lines in `titleGeneration.ts`.
  Whichever of the two merges second, I will rebase it — say the word and I will push
  the rebase.
- **#118** (merged) supplied the word-boundary truncation this PR promotes to shared
  code; behaviour for the fallback path is unchanged apart from the wider budget.
- Independent of #119, #121, #122.

### Verification

Automated commands run:

```text
npm run lint
npm run typecheck
npx jest --selectProjects unit
npm run review:source
npm run review:css
npm run review:deps
npm run build:release
```

Manual verification: Obsidian on Windows 11, running a local build. Newly generated
titles use the wider budget and are cut on a word boundary; the previously truncated
`... error Cl...` title regenerates into a complete one.

Fixes #125

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,6 +110,7 @@ The project requires Obsidian 1.13.0 or newer and uses the declarative settings 
 |------|-------|
 | `.grimoire/grimoire-settings.json` | Shared Grimoire app settings plus provider-specific configuration |
 | `.grimoire/sessions/*.meta.json` | Provider-neutral session metadata |
+| `.grimoire/attachments/<sha256>.<ext>` | Image attachment bytes, addressed by content and shared by every provider |
 | `.grimoire/logs/YYYY-MM-DD.jsonl` | Optional sanitized debug logs, written only when Advanced debug logging is enabled |
 | `.grimoire/mcp/<provider>.json` | Grimoire-owned MCP servers injected into ACP sessions for OpenCode, Grok Build, MiMoCode, Kimi Code, Qwen Code, and Gemini CLI |
 | `.grimoire/claude/statusline-usage.json` | Claude Code status-line usage snapshot used to hydrate plan-limit indicators |
--- a/src/app/storage/SharedStorageService.ts
+++ b/src/app/storage/SharedStorageService.ts
@@ -1,6 +1,7 @@
 import type { Plugin } from 'obsidian';
 import { Notice } from 'obsidian';
 
+import { AttachmentStore } from '../../core/attachments/AttachmentStore';
 import { SESSIONS_PATH, SessionStorage } from '../../core/bootstrap/SessionStorage';
 import type { SharedAppStorage } from '../../core/bootstrap/storage';
 import { GRIMOIRE_STORAGE_PATH } from '../../core/bootstrap/StoragePaths';
@@ -63,6 +64,7 @@ function normalizeDraftSettings(value: unknown): Record<string, unknown> | undef
 export class SharedStorageService implements SharedAppStorage {
   readonly grimoireSettings: GrimoireSettingsStorage;
   readonly sessions: SessionStorage;
+  readonly attachments: AttachmentStore;
 
   private adapter: VaultFileAdapter;
   private plugin: Plugin;
@@ -72,6 +74,7 @@ export class SharedStorageService implements SharedAppStorage {
     this.adapter = new VaultFileAdapter(plugin.app);
     this.grimoireSettings = new GrimoireSettingsStorage(this.adapter);
     this.sessions = new SessionStorage(this.adapter);
+    this.attachments = new AttachmentStore(this.adapter);
   }
 
   async initialize(): Promise<{ grimoire: Record<string, unknown> }> {
--- a/src/core/AGENTS.md
+++ b/src/core/AGENTS.md
@@ -6,6 +6,8 @@
 
 - `runtime/` defines neutral chat runtime contracts such as `ChatRuntime`, prepared turns, query options, approvals, and session updates.
 - `providers/` defines registries, capabilities, model routing, environment, settings projection, and workspace-service contracts.
+- `attachments/` owns the content-addressed store for image attachment bytes, plus the scaling
+  policy applied before they are stored. Provider-neutral: every provider reads the same files.
 - `bootstrap/` owns provider-neutral session metadata storage and shared app-storage contracts.
 - `debug/` owns sanitized file logging. Debug logs must be opt-in and must not include prompts, answers, note content, paths, environment values, or secrets.
 - `mcp/` owns provider-neutral MCP config parsing, testing, and coordination.
--- a/src/core/attachments/AttachmentStore.ts
+++ b/src/core/attachments/AttachmentStore.ts
@@ -1,0 +1,111 @@
+import { createHash } from 'crypto';
+
+import type { ImageMediaType } from '../types';
+
+/** Grimoire-owned vault folder holding attachment bytes, addressed by content. */
+export const ATTACHMENTS_FOLDER = '.grimoire/attachments';
+
+const EXTENSIONS: Record<ImageMediaType, string> = {
+  'image/png': 'png',
+  'image/jpeg': 'jpg',
+  'image/gif': 'gif',
+  'image/webp': 'webp',
+};
+
+/** The slice of the vault adapter the store needs. */
+export interface AttachmentFileStore {
+  exists(path: string): Promise<boolean>;
+  readBinary(path: string): Promise<ArrayBuffer>;
+  writeBinary(path: string, data: ArrayBuffer): Promise<void>;
+  delete(path: string): Promise<void>;
+  listFiles(folder: string): Promise<string[]>;
+  getResourcePath(path: string): string;
+}
+
+export interface StoredAttachment {
+  hash: string;
+  mediaType: ImageMediaType;
+  size: number;
+}
+
+export function attachmentFileName(hash: string, mediaType: ImageMediaType): string {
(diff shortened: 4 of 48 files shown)


sandsaber:
Pushed review fixes to this branch — details in the commit message.

 Two producers were still cutting the same text differently, because the shared helper took the trailing trim as a parameter and only the fallback path passed one that strips punctuation:

```
generated  "Fixing the parser, the tokenizer,..."     fallback  "…tokenizer..."
generated  "Diagnose the 500 error -..."              fallback  "…error..."
```

One budget is half of what sharing a truncation is for; the other half is that a title does not end differently depending on who cut it. `trimTitleNoise` moved into the shared module as the only behaviour and the parameter is gone, so the API got smaller too.

The prompt no longer asks for the budget. It is a safety net against mutilation, not a target: asking for 100 characters made every generated title twice as long (measured: fallback 98, model answer 97), and a tab or a history row ellipsises it anyway. `TITLE_PROMPT_TARGET_LENGTH` keeps the ask at 50 while the budget stays 100, so the reported `... error Cl...` case survives whole without lengthening everything else. If you would rather ask for 100, it is one constant.

`buildForkTitle` now cuts through the shared helper as well — it was the one producer the change did not reach, still slicing mid-word with its own `…`.

I verified the two claims that mattered: the extracted truncation is byte-identical to the one #118 introduced, and the adjusted fallback tests are not weakened — same assertions, `maxLength: 50` passed explicitly.

Full gate on the branch merged with current `main`: `npm test`, `typecheck`, `lint` all clean.

PR #128, opened 2026-09-05
fix(claude): stop reopening a conversation from duplicating every message

### LINKED ISSUE #127: Reopening a Claude conversation duplicates every message
## What happens

Reopening a Claude conversation whose SDK transcript is still on disk appends a
second copy of every message. The UI shows the whole exchange twice, and the
duplicated list is then saved back into `.grimoire/sessions/<id>.meta.json`.

Claude only: it is the only provider that merges an external transcript into
`conversation.messages`.

The duplication is stable at ×2 rather than growing, which is why it reads as a
long conversation rather than a bug.

## Reproduction

1. Send a message in a Claude conversation and wait for the answer.
2. Switch to another conversation and back (or restart Obsidian and reopen it).
3. Every message appears twice.

## Cause

`dedupeMessages` in `src/providers/claude/history/ClaudeConversationHistoryService.ts:166`
keys only on `message.id`:

```ts
for (const message of messages) {
  if (seen.has(message.id)) continue;
  seen.add(message.id);
  result.push(message);
}
```

It is fed two sets whose ids come from different namespaces:

- `conversation.messages`, with Grimoire ids of the form `msg-<timestamp>-<rand>`;
- the parsed transcript, whose ids are `sdkMsg.uuid` (`sdkMessageParsing.ts:238`).

They never collide, so the merge at
`ClaudeConversationHistoryService.ts:454-457` appends both copies.

It stops at ×2 because the saved list then contains the uuid-keyed copies, which
do dedupe against the transcript on the next hydration.

## The identifiers needed are already stored

`ChatMessage` already carries `userMessageId` / `assistantMessageId`, filled in
`InputController.ts:635-636` from `agentService.consumeTurnMetadata()`, which
`ClaudeChatRuntime` populates with the transcript uuid. Existing metadata in a
vault shows both halves of each pair sharing a key:

```
user      id=msg-1786188398292-x8   userMessageId=f428a7aa-c360-4b0f-9
user      id=f428a7aa-c360-4b0f-9   userMessageId=f428a7aa-c360-4b0f-9

assistant id=msg-1786188398314-wv   assistantMessageId=ecb3966c-cfdb-4790-8
assistant id=b9d8d624-1d51-4f46-b   assistantMessageId=ecb3966c-cfdb-4790-8
```

A user duplicate's `id` equals the original's `userMessageId`; assistant copies
share an `assistantMessageId`.

## Fix

Dedupe on the union of `{id, userMessageId, assistantMessageId}` instead of `id`
alone, keeping the first occurrence so the Grimoire-side message wins - it holds
the id the UI uses for rewind and fork, and the attachment-store hash on its
images.

Content comparison is not the answer: repeated prompts ("what is this?") are
legitimately distinct messages.

## Already-duplicated conversations

They repair themselves. The stored `msg-…` copies carry the matching uuid, so the
next hydration collapses each pair. No migration pass is needed.

A message that never received turn metadata - an interrupted turn, say - has
neither identifier and will not collapse. It is not made worse.

## Not covered by tests

No test opens a Claude conversation with a live transcript on disk, which is why
this survived. The fix should add one.

Found while testing #126 on a live vault; unrelated to that change - vault
metadata shows the same duplication in conversations from 1, 8, 24, 25 and 31
August, none of which involve images.

sandsaber:
Closes #127

## Cause

`dedupeMessages` (`ClaudeConversationHistoryService.ts:166`) keyed only on `message.id`, and the merge feeds it two sets whose ids come from different namespaces — the conversation's own `msg-<timestamp>-<rand>` ids and the transcript's `sdkMsg.uuid`. They never collide, so every hydration appended a second copy of the whole exchange, which was then saved back into `.grimoire/sessions/<id>.meta.json`.

It stopped at ×2 rather than growing, because the saved list then held the uuid-keyed copies, which do dedupe against the transcript on the next pass. That is why it read as a long conversation rather than a bug.

Claude only: it is the only provider that merges an external transcript into `conversation.messages`.

## Fix

The identifiers needed were already stored. `ChatMessage.userMessageId` / `assistantMessageId` are filled in `InputController.ts:635-636` from the turn metadata Claude's runtime records, and vault data shows both halves of each pair sharing a key:

```
user      id=msg-1786188398292-x8   userMessageId=f428a7aa-c360-4b0f-9
user      id=f428a7aa-c360-4b0f-9   userMessageId=f428a7aa-c360-4b0f-9

assistant id=msg-1786188398314-wv   assistantMessageId=ecb3966c-cfdb-4790-8
assistant id=b9d8d624-1d51-4f46-b   assistantMessageId=ecb3966c-cfdb-4790-8
```

Deduping on the union of `{id, userMessageId, assistantMessageId}` collapses each pair. The first copy wins, which keeps the conversation's own message — the UI refers to it by id for rewind and fork, and its attachments carry the attachment-store hash.

Matching on content was not an option: a repeated prompt ("what is this?") is a legitimately distinct message.

## Already-duplicated conversations

They repair themselves on the next open, because both stored copies share an identifier. No migration pass.

A message that never received turn metadata — an interrupted turn — has neither identifier and is left alone rather than merged into a neighbour.

## Tests

No test opened a Claude conversation against a transcript, which is how this survived since at least 1 August. Six now do: the duplicate is not appended, the conversation's copy is the survivor (with its attachment hash intact), an already-duplicated conversation collapses, genuinely new transcript messages are still added, two identical prompts stay apart, and a message with no SDK identifier is left alone.

Verified they fail without the fix: **4 of 6 fail** on the unpatched service, all 6 pass with it.

`npm test` (425 suites, 7750 tests), `typecheck`, `lint` — clean.

### DIFF
--- a/src/providers/claude/history/ClaudeConversationHistoryService.ts
+++ b/src/providers/claude/history/ClaudeConversationHistoryService.ts
@@ -163,13 +163,47 @@ function ensureTaskToolCall(
   return taskToolCall;
 }
 
+/**
+ * The identifiers under which a message can be recognised again.
+ *
+ * A message reaches the merge from two namespaces: Grimoire's own `msg-…` ids,
+ * and the transcript's uuids. The same exchange therefore arrives under two
+ * different `id`s, and only the SDK-side identifiers tie them together - a user
+ * message's transcript uuid is its `userMessageId`, and both copies of an
+ * assistant message carry the same `assistantMessageId`.
+ */
+function messageIdentities(message: ChatMessage): string[] {
+  const identities = [message.id];
+  if (message.userMessageId) {
+    identities.push(message.userMessageId);
+  }
+  if (message.assistantMessageId) {
+    identities.push(message.assistantMessageId);
+  }
+
+  return identities;
+}
+
+/**
+ * Keeps the first copy of each message.
+ *
+ * Order matters: the conversation's own messages come first, so the surviving
+ * copy is the one the UI already refers to by id for rewind and fork, and the
+ * one whose attachments point at the store.
+ *
+ * Matching on identifiers rather than on content is deliberate - a repeated
+ * prompt is a legitimately distinct message, not a duplicate.
+ */
 function dedupeMessages(messages: ChatMessage[]): ChatMessage[] {
   const seen = new Set<string>();
   const result: ChatMessage[] = [];
 
   for (const message of messages) {
-    if (seen.has(message.id)) continue;
-    seen.add(message.id);
+    const identities = messageIdentities(message);
+    if (identities.some(identity => seen.has(identity))) continue;
+    for (const identity of identities) {
+      seen.add(identity);
+    }
     result.push(message);
   }
 
--- a/tests/unit/providers/claude/history/ClaudeConversationHistoryService.test.ts
+++ b/tests/unit/providers/claude/history/ClaudeConversationHistoryService.test.ts
@@ -1,0 +1,156 @@
+import type { ChatMessage, Conversation } from '@/core/types';
+import { ClaudeConversationHistoryService } from '@/providers/claude/history/ClaudeConversationHistoryService';
+
+jest.mock('@/providers/claude/history/ClaudeHistoryStore', () => ({
+  deleteSDKSession: jest.fn(),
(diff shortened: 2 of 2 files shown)


sandsaber:
@greenkms thanks for your hard work, effort, and contribution!

Issue #141, opened 2026-09-07
Antigravity turn lifecycle: three 1.3.2 quantities kept their value but changed meaning in 2.0


greenkms:
## Summary

Four defects in the Antigravity turn lifecycle share one cause: a quantity carried over from 1.3.2 kept its value while the code around it changed what the value meant. Each one alone reads as an unrelated bug; together they describe a single review question worth asking of the rest of the 2.0 kernel.

Reported separately as a user-visible symptom in #139. This issue is the underlying class.

## 1. A parse buffer size became an answer budget

In 1.3.2 `OUTPUT_BUFFER_LIMIT = 64_000` was the size of a **sliding** buffer: the code kept the last 64 KB of output and discarded the head (`.slice(-64_000)`). Carrying the same number into 2.0, where the limit became a **cumulative** budget for the turn, silently turned "keep the last 64 KB" into "kill any turn that says more than 64 KB".

This is not a theoretical ceiling. A trivial `agy` probe already writes roughly 30 KB, so ordinary turns died with `The response exceeded its size limit` and stored an empty answer.

The value is now a memory ceiling rather than a budget on what the agent is permitted to say.

## 2. A pipe deadline was dropped, then reintroduced as an unbounded wait

1.3.2 gave the streams a grace period after process exit (`PRINT_DRAIN_GRACE_MS = 2_000`) and then forced them shut, because an orphaned grandchild holding the pipes would otherwise hold the whole run hostage.

2.0 waits on the streams with no deadline. When a grandchild keeps the pipes open after `agy` exits, the answer is in, the process is gone, and the turn never settles.

## 3. Process exit was treated as the end of the turn

`agy` does not exit after producing an answer — verified with `Get-Process agy` against a live run, where the process stayed resident after the response was complete. Waiting for it to leave leaves the tab spinning on an answer it already has.

In stream-json the turn is over when the last `result` frame arrives, not when the process does. The `text_delta` frames sum exactly to `result.response`, so the frame is both the terminal signal and the complete answer.

## 4. Process liveness was treated as turn liveness

The most subtle of the four, and the one that survived the first three fixes.

`agy` emits frames on step transitions rather than continuously, so a single long tool call can keep both pipes quiet — one measured at about five minutes in the wild. To tell that apart from a hang, the runner polls the run's log file and reports growth as activity, which re-arms the backend inactivity timeout.

[...]

PR #142, opened 2026-09-07
fix(antigravity): end a turn on its own evidence, not on the CLI process

### LINKED ISSUE #139: Antigravity turns remain running after the final answer is visible and only finish after Escape
## Summary

An Antigravity turn can finish producing a complete, visible assistant answer while Grimoire continues to show the run as active indefinitely. The user must press Escape to finish the turn.

This still occurs after installing a local patch which stops waiting for the `agy` process to exit and instead attempts to complete the turn when the provider emits its terminal `result` frame.

The remaining failure therefore appears to be a distinct completion-seam problem: in at least one real `agy` output sequence, the answer becomes visibly complete but Grimoire does not observe or accept the terminal evidence required to settle the run.

The exact missing transition is still under investigation. This report does not assume whether the `result` frame is absent, delayed, malformed, lost by the parser, or rejected by correlation logic.

## Environment

- OS: Windows
- Host: Obsidian desktop
- Grimoire manifest version: 1.3.2
- Tested source baseline: upstream commit `65dc6775`
- Locally installed diagnostic build includes the focused Antigravity completion patch and later unrelated chat-restoration fixes through commit `074f15af`
- Provider: Antigravity CLI (`agy`) in stream-json mode

## Steps to reproduce

1. Open an Antigravity conversation in Grimoire.
2. Send a request which produces a normal assistant response.
3. Wait until the assistant answer appears complete in the conversation.
4. Observe that the run remains in the active/streaming state.
5. Continue waiting for several minutes.
6. Press Escape.

The issue has reproduced repeatedly, although the delay before manual cancellation varies.

## Actual behavior

- The assistant response is already visible and appears complete.
- Grimoire continues to treat the execution run as `running`.
- The provider process remains alive.
- The run does not settle on its own within several minutes.
- Pressing Escape changes the run to `cancelled / cancellation-confirmed`.

Recent persisted run evidence includes multiple accepted Antigravity runs for the same conversation which remained active for approximately five to six minutes and were then cancelled manually.

No claim is made that the provider process itself should exit after every response. Earlier observation established that `agy` may remain alive after producing an answer, so process exit is not a valid completion boundary by itself.

## Expected behavior

Once the provider has completed the response, Grimoire should settle the execution run without requiring Escape.

The completion condition should be based on provider-native terminal evidence and should work for every valid `agy` stream sequence that represents a completed answer.

If terminal evidence never arrives, Grimoire should not wait forever. It should eventually expose a bounded and diagnostically useful failure which distinguishes at least:

- complete answer received but terminal frame missing;
- terminal frame received but not parsed;
- terminal frame parsed but not correlated with the active run;
- provider still actively producing output;
- provider stopped producing output without confirming completion.

## Prior fix and why this is separate

The original Antigravity completion bug was caused by treating process exit as the end of the turn. Real observation showed that `agy` can emit the answer and remain alive.

A focused local patch changed the completion seam to the provider's terminal `result` frame and then requests graceful CLI shutdown. That patch has focused regression coverage for a recognized result-frame sequence.

The current reproduction occurs with that patch installed. It therefore should not be closed as the already-known “process does not exit” defect. Either the real provider sequence does not contain the expected terminal frame, or the frame does not reach the state transition covered by the existing test.

## Investigation boundary

The next diagnostic pass should record only sanitized protocol structure and timing, not prompts or response text. For each provider frame it should capture:

- frame type;
- timestamp or monotonic offset;
- whether parsing succeeded;
- correlation identifiers in redacted or synthetic form;
- whether the frame was accepted for the active run;
- the run state before and after processing;
- whether graceful termination was requested;
- whether any output arrived after the apparently final assistant text.

This is needed to distinguish a provider protocol variant from a parser, buffering, or run-correlation bug.

## Impact

- Every affected Antigravity turn requires manual cancellation.
- The UI continues to indicate work after the answer is already available.
- Users cannot reliably distinguish a completed answer from a still-running agent.
- Escape is used as a normal completion mechanism even though it semantically represents cancellation.
- Post-turn persistence and subsequent conversation behavior depend on a manual intervention.
- Repeated long-lived provider processes and runs can complicate shutdown and recovery.

The observed transcript remains available; this issue is about execution completion, not confirmed transcript loss.

## Security and privacy

No permission or authorization change is requested.

Provider frames, model output, paths, session identifiers, and process output must remain untrusted. Diagnostics and fixtures must not contain prompts, note contents, credentials, local vault paths, complete real session identifiers, or unsanitized provider payloads.

Any fallback completion rule must be conservative. Visible text alone is not sufficient proof that tools, filesystem operations, or other provider-side work have finished.

## Suggested acceptance criteria

- [ ] A real Antigravity response which is complete at the provider protocol level settles without Escape.
- [ ] The regression test covers the exact structured frame sequence observed in the failing runtime.
- [ ] The test fails on the current implementation before the fix.
- [ ] A recognized terminal `result` frame still completes the run exactly once.
- [ ] A provider process remaining alive does not keep an already terminal run open.
- [ ] Missing terminal evidence cannot leave the run waiting indefinitely.
- [ ] Any fallback timeout distinguishes inactivity from absolute runtime and does not terminate an actively working turn.
- [ ] Cancellation retains precedence when the user presses Escape before terminal evidence is confirmed.
- [ ] Late or duplicate terminal frames cannot settle a later run.
- [ ] Sanitized diagnostics identify whether a frame was absent, unparsed, or rejected without logging response content.
- [ ] Manual verification is performed in Obsidian with the real Antigravity provider on Windows.

## Patch status

We are actively working on a focused patch for this remaining completion failure.

The planned sequence is:

1. capture a sanitized structural trace of one reproduced hanging turn;
2. add a failing regression test using the observed structured frame sequence;
3. fix the smallest provider-owned completion boundary responsible for the missing terminal transition;
4. preserve cancellation, timeout, and process-ownership semantics;
5. run the focused Antigravity tests, typecheck, lint, the complete unit comparison, and the release build;
6. install the verified release artifacts in a test vault;
7. confirm that several real Antigravity turns complete without Escape after their final answers appear.

The earlier result-frame patch is not being presented as a complete fix for this newly reproduced sequence. No new patch or PR is ready for review yet.

greenkms:
Fixes #139. Background and the shared cause behind all four commits: see the turn-lifecycle issue.

Four commits on one seam, in the order the defects were found — each became visible only once the one before it was fixed.

## 1. `stop killing a turn that says more than a parse buffer held`

`OUTPUT_BUFFER_LIMIT = 64_000` was the size of a *sliding* buffer in 1.3.2: keep the last 64 KB, discard the head. In 2.0 the same number became a cumulative budget for the turn, which turns "keep the last 64 KB" into "kill any turn that says more than 64 KB". A trivial `agy` probe already writes ~30 KB, so real turns died with `The response exceeded its size limit` and stored an empty answer.

Now a memory ceiling rather than a limit on what the agent may say.

## 2. `give the pipes a deadline once the process has gone`

1.3.2 gave the streams a grace period after exit and then forced them shut, because an orphaned grandchild holding the pipes would otherwise hold the run hostage. 2.0 waited on them with no deadline: answer in, process gone, tab spinning on a pipe nobody will close.

## 3. `end the turn on the result frame, not on the process leaving`

`agy` does not exit after answering — verified against a live run where the process stayed resident with a complete response. In stream-json the turn ends at the last `result` frame; the `text_delta` frames sum exactly to `result.response`, so the frame is both the terminal signal and the whole answer. A CLI that overstays its own answer is then asked to leave.

## 4. `stop a resident CLI's own log from re-arming the turn`

The one that survived the first three, and the direct cause of #139.

`agy` emits frames on step transitions, so a long tool call keeps both pipes quiet — one measured at about five minutes. To tell that apart from a hang, the runner polls the run's log file and reports growth as activity, re-arming the backend inactivity timeout.

But the log belongs to the CLI, not to the turn. A resident `agy` keeps appending after the answer is delivered, so unconditional reporting re-arms the timeout on every poll. A turn whose `result` frame never arrives then has no boundary left but the 30-minute absolute ceiling. Observed: turns active for five to eleven minutes with a complete visible answer, ending only on Escape. Every such run recorded `wasInterrupted: true` and `planCompleted: false` — none ended on its own.

`agy` brackets every tool call with an `ACTIVE`/`DONE` pair sharing one step id, so the set of calls started and not yet finished is exact evidence that the CLI still owes this turn work. Log growth counts while a call is out; after the last `DONE`, silence is silence.

Deliberately not done here: nothing ends a turn because text became visible, and cancellation and process ownership are untouched. A run with no frames to read keeps the unconditional watch it has always had, and the log size is still recorded when it is not reported, so growth during a quiet stretch cannot arrive later as one jump and read as the run waking up.

## Tests

Each commit carries a regression test that fails before it.

The fourth also carries the complementary case, so the behaviour it narrows cannot be removed by accident: while a tool call is open, log growth must still count as liveness.

One existing test was found to pass for the wrong reason and is now backed by a stricter sibling: its fake stdout *ended* after the result frame, so the drain resolved and the wait was released by the pipe rather than by the frame. The added case keeps the pipe open, which is what a resident CLI does.

## Verification

- focused Antigravity suites: 17 passed, 185 tests passed
- `npx tsc --noEmit`: exit 0
- `npx eslint "src/**/*.ts" "tests/**/*.ts" --max-warnings=0`: exit 0
- full unit run: 23 failed, 9001 passed — the same pre-existing Windows failures as the `65dc6775` baseline, no delta
- `npm run build:release`: exit 0, source, CSS and dependency review gates passed

Issue: #141

### DIFF
--- a/src/providers/antigravity/execution/AntigravityExecutionComposition.ts
+++ b/src/providers/antigravity/execution/AntigravityExecutionComposition.ts
@@ -54,7 +54,23 @@ import {
 } from '../runtime/AntigravityImageAttachments';
 
 /** Combined stdout, stderr, and recovered-transcript ceiling for one run. */
-const OUTPUT_BYTE_LIMIT = 64_000;
+/**
+ * How much one print turn may say before the run is stopped.
+ *
+ * A ceiling on memory, not a budget for an answer: the runner keeps the whole
+ * stream so the transcript can be parsed, so something has to bound it.
+ *
+ * The number matters. It used to be `OUTPUT_BUFFER_LIMIT = 64_000` — the size
+ * of a *sliding* parse buffer (`.slice(-OUTPUT_BUFFER_LIMIT)`), which bounded
+ * memory while letting a turn print without end. Carried over to a cumulative
+ * budget, the same 64 KB stopped meaning "keep the last 64 KB" and started
+ * meaning "kill any turn that says more than 64 KB". A trivial agy probe
+ * already writes ~30 KB of stream-json, so real turns were terminated with
+ * `output-limit` and stored an empty answer under the question that asked for
+ * it.
+ */
+export const ANTIGRAVITY_OUTPUT_BYTE_LIMIT = 16 * 1024 * 1024;
+const OUTPUT_BYTE_LIMIT = ANTIGRAVITY_OUTPUT_BYTE_LIMIT;
 
 /**
  * Antigravity chat execution, assembled from the running plugin.
--- a/src/providers/antigravity/runtime/AntigravityPrintProcessRunner.ts
+++ b/src/providers/antigravity/runtime/AntigravityPrintProcessRunner.ts
@@ -67,6 +67,14 @@ export interface AntigravityPrintProcessRunnerOptions {
   readonly logSize?: (logFilePath: string) => Promise<number>;
   /** Combined stdout, stderr, and recovered-result byte ceiling. */
   readonly outputByteLimit?: number;
+  /**
+   * How long the pipes may keep talking after the process has gone.
+   *
+   * Not a timeout for the turn: the turn is over when `agy` exits. This only
+   * bounds the wait for buffered output that an orphaned grandchild may be
+   * holding open, so a finished run cannot hang on a pipe nobody will close.
+   */
+  readonly drainGraceMs?: number;
   readonly createLogPath?: () => string;
   readonly recoverTranscript?: (
     logFilePath: string,
@@ -182,6 +190,10 @@ export class AntigravityPrintProcessRunner implements AntigravityProcessRunner {
     };
   }
 
+  private drainGraceMs(): number {
+    return this.options.drainGraceMs ?? 2_000;
+  }
+
   private async observeCompletion(
     child: AntigravityManagedChildProcess,
     invocation: AntigravityInvocation,
@@ -194,13 +206,22 @@ export class AntigravityPrintProcessRunner implements AntigravityProcessRunner {
     const stderr = new LimitedBytes(outputLimit);
     const onActivity = hooks.onActivity;
     try {
-      const [exit] = await Promise.all([
-        child.exited,
+      // The process leaving is the event; the pipes closing is not. A grandchild
+      // that outlives `agy` keeps them open, and waiting on them with no
+      // deadline holds the run hostage: the answer is in, the process is gone,
+      // and the tab spins forever. So the streams get a grace period *after*
+      // exit and the outcome is built from whatever arrived by then — which is
+      // what 1.3.2 did by forcing the streams shut once `close` lagged `exit`.
+      const drained = Promise.all([
         parser
           ? consumeFrames(child.stdout, parser, outputLimit, onActivity)
           : consume(child.stdout, stdout, onActivity),
         consume(child.stderr, stderr, onActivity),
-      ]);
+        // A pipe that fails after the process already left says nothing about
+        // the turn, and must not replace its outcome with a rejection.
+      ]).catch(() => undefined);
+      const exit = await child.exited;
+      await Promise.race([drained, delay(this.drainGraceMs())]);
       // **The frame, not the pipe.** In stream-json the answer is one field of
       // the last `result` frame, and the frames around it are progress this
       // turn has already published. Accumulating the pipe instead would spend
@@ -335,6 +356,10 @@ async function consumeFrames(
     }
     parser.write(decoder.decode(chunk, { stream: true }));
   }
+}
+
+function delay(ms: number): Promise<void> {
+  return new Promise(resolve => { window.setTimeout(resolve, ms); });
 }
 
 function removeLog(logFilePath: string): Promise<void> {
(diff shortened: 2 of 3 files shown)


greenkms:
Closing in favour of a new PR, which supersedes this branch.

The first three commits here are good and carry over unchanged. The fourth — narrowing the log-growth liveness signal to periods when a tool call is open — was based on an explanation that later evidence disproved, and I would rather withdraw it than have it merged on reasoning that did not hold.

What the fourth commit assumed: that `agy` keeps appending to its own log after answering, so reporting that growth as activity re-armed the inactivity timeout indefinitely and left the turn with no boundary but the absolute ceiling.

What was actually wrong: the wait itself. It was a race between the process exiting and an `all` of *both* pipe consumers. The stdout consumer returns the moment the `result` frame is read, but a resident `agy` holds stderr open too, so the `all` stays pending — and the exit never comes either. Both arms of the race end up waiting on a CLI that has already answered.

The measurement that settled it, from a real hung turn: `init`, two `step_update` and `result` all arrived within 7.6 seconds, `hasResult` true — and the turn still hung for eleven minutes. So no terminal frame was missing and no timer was being re-armed; the turn simply had nothing left that could end it.

That also makes the fourth commit the wrong shape on its own terms: it ended a hung turn by letting a ten-minute timeout expire, which is a workaround, not a fix. The replacement uses no timeout at all — the stdout side becomes its own arm of the race, so the answer arriving is reason enough to stop waiting.

The new PR keeps the three commits from here and adds three more: unconditional EOF on stdin (1.3.2 closed unconditionally; 2.0 made the close wait on a write callback that may never fire), retention of the run log on failure plus a sanitised completion record — which is what found the real defect on the first hung turn after it shipped — and the fix above.

Sorry for the noise on this one.

Issue #137, opened 2026-09-06
Codex conversations can become blocked after restart when duplicate execution sessions contend for the same native thread


greenkms:
## Summary

After restarting Obsidian, sending a message in an existing Codex conversation can fail immediately with:

> ⚠️ Blocked: Grimoire could not establish what this run changed.

The visible message reports an indeterminate execution with unknown side effects. However, the observed provider failure happens while Grimoire is preparing or resuming the native Codex thread, before `turn/start` is sent.

The reproduced failure involves two Grimoire execution-session records associated with the same conversation:

- one execution session retains the conversation's native Codex thread reference;
- another execution session exists for the same conversation without a `nativeSessionRef`;
- the new run is dispatched through the latter session;
- that session attempts to resume a native thread which is already owned or loaded elsewhere;
- Codex rejects the resume with an active-writer conflict;
- Grimoire classifies the result as `indeterminate / effects-unknown`.

Creating a completely new Codex conversation works because it receives a new native thread and does not encounter the conflicting restored state.

## Environment

- OS: Windows
- Host: Obsidian desktop
- Grimoire manifest version: 1.3.2
- Tested source baseline: upstream commit `65dc6775`
- Locally installed diagnostic build also contains unrelated, focused chat-restoration patches through commit `074f15af`
- Provider: Codex through `codex app-server`

The local patches are not believed to cause this Codex failure. The same Codex resume problem was observed before the restored-tab binding patch was installed.

## Steps to reproduce

1. Open an existing Codex conversation which already has a persisted native Codex thread.
2. Complete one or more turns successfully.
3. Restart Obsidian normally.
4. Reopen or use the restored Codex conversation.
5. Send another message.

## Actual behavior

The run terminates almost immediately with:

> Grimoire could not establish what this run changed.

Observed control-plane state:

- `dispatchState: accepted`
- terminal state: `indeterminate`
- terminal reason: `effects-unknown`
- no `nativeRunRef`
- no assistant content
- completion approximately 0.3 seconds after dispatch

In the latest reproduction, one conversation had two execution-session records:

1. an execution session with the persisted native thread reference;
2. an execution session for the same conversation without that reference.

The failing run used the second session.

A direct read-only protocol reproduction previously exposed the underlying Codex error as an active-writer conflict during `thread/resume`. No `turn/start` was observed for the failed attempt.

## Expected behavior

After restart, an existing Codex conversation should have one authoritative execution-session/native-thread binding.

Grimoire should either:

[...]

PR #154, opened 2026-09-07
fix(providers): give the ACP startup handshake a budget of its own

### LINKED ISSUE #153: A new ACP tab fails its first turn and blames a saved session that never existed
## Summary

A brand-new conversation on Gemini, OpenCode or MiMoCode fails its first turn with:

> <Provider> could not start this turn. If this conversation was resumed from a saved session, that session may no longer exist — starting a new chat will create one.

The advice cannot be acted on. The conversation was created seconds earlier and has no saved session — its stored record has `sessionId: null` — and the new chat it recommends fails identically. Grok, Claude, Codex and Antigravity are unaffected.

Two separate defects meet here.

**The budget is the wrong one.** `ManagedAcpExecutionBackend` wraps `client.initialize()` in `controlTimeoutMs ?? 2_000`. That is a *control* budget: two seconds is a fair wait for `session/cancel` on a process already running, and it is not a fair wait for a CLI that is still booting. None of the six ACP compositions overrides it.

Measured against the shipped agents — spawn plus a real `initialize` handshake at `protocolVersion: 1`, three runs each, idle machine, no MCP servers:

| CLI | `initialize` | share of the 2 000 ms budget | tab |
|---|---|---|---|
| `grok agent stdio` | 403 / 432 / 624 ms | ~25 % | works |
| `opencode acp` | 1448 / 1474 / 1481 ms | ~74 % | fails |
| `mimo acp` | 1414 / 1440 / 1472 ms | ~74 % | fails |
| `gemini --acp` | 2073 / 2124 / 2153 ms | **over, every run** | fails |

Spawn itself is 6–13 ms in every case, so the cost is the handshake. Gemini never fits, even in the ideal conditions above. OpenCode and MiMo spend three quarters of the budget before Obsidian's own renderer — `window.setTimeout` on a busy event loop, with the full MCP server list — is accounted for. The cut-off falls exactly between Grok and the other three, which is the split the affected tabs actually show.

Note this is a latent defect, not a recent regression: it is present on current `main`, and it surfaced as the CLIs grew slower and Gemini crossed the line.

**The reason is then discarded.** The timeout raises a plain `ExecutionDispatchError`, and only a `JsonRpcErrorResponse` is ever carried to the tab as a turn refusal. So the run reaches the surface with no words at all, `describeFailure` falls through to the wordless `pre-dispatch-rejected` branch, and `describeAcpSessionOpenFailure` explains the silence with the only sentence it has — the one about a saved session.

This is what made the defect hard to see from the outside: the message points away from the cause and names a remedy that cannot work.

## Steps to reproduce

1. Configure Gemini, OpenCode or MiMoCode with a valid CLI path.
2. Open a **new** tab for that provider — no resumed conversation.
3. Send any short prompt, for example `just say ok`.

## Expected

The turn starts. A CLI that is merely slow to come up is not a CLI that refused, and a startup that does fail should say what failed.

## Actual

The turn is rejected before dispatch, the conversation is left empty with `sessionId: null`, and the tab reports that a saved session may no longer exist — for a conversation that never had one.

## Notes

A fix is prepared: `initialize` gets its own `startupTimeoutMs` (30 s by default), and a startup failure becomes `AcpStartupError`, travelling as a turn refusal with origin `startup` so the tab can say what happened instead of offering a new chat. Verified live on all three affected providers after install; the named delta of failing tests against `main` is empty in both directions.

greenkms:
Fixes #153 

## What was wrong

A brand-new conversation on Gemini, OpenCode or MiMoCode failed its first turn and was told that the saved session it had been resumed from may no longer exist. It had never been resumed from anything: the record shows `sessionId: null`, the conversation was seconds old, and the new chat the message recommends fails identically.

Two defects met in one place.

**The budget was the wrong one.** `initialize` was wrapped in `controlTimeoutMs ?? 2_000`. That is a *control* budget — two seconds is a fair wait for `session/cancel` on a process already running, and it is not a fair wait for a CLI that is still booting. None of the six ACP compositions overrode it.

Measured on the shipped agents (spawn plus a real handshake at `protocolVersion: 1`, three runs each, idle machine, no MCP servers):

| CLI | `initialize` | share of the 2 000 ms budget | tab |
|---|---|---|---|
| `grok agent stdio` | 403 / 432 / 624 ms | ~25 % | works |
| `opencode acp` | 1448 / 1474 / 1481 ms | ~74 % | fails |
| `mimo acp` | 1414 / 1440 / 1472 ms | ~74 % | fails |
| `gemini --acp` | 2073 / 2124 / 2153 ms | over, every run | fails |

Spawn is 6–13 ms throughout, so the cost is the handshake itself. Gemini never fits even in those ideal conditions; OpenCode and MiMo spend three quarters of the budget before Obsidian's own renderer — `window.setTimeout` on a busy event loop, with the full MCP server list — is accounted for. The cut-off falls exactly between Grok and the other three, which is the split the affected tabs showed.

`initialize` now has `startupTimeoutMs`, defaulting to 30 s.

**The reason was then thrown away.** The timeout raised a plain `ExecutionDispatchError`, and only a `JsonRpcErrorResponse` was ever carried to the tab. The run reached the surface with no words at all, so the composition explained the silence with the only sentence it had. A startup failure is now `AcpStartupError`; it travels as a turn refusal with origin `startup` and gets a sentence that names what happened and does not offer a new chat, because a new chat fails the same way.

`presentTurnRefusal` is called in `prepare`, where both the run and the error are in scope: the client is ensured before the session binding is opened, so the existing `catch` in `openSessionBinding` never saw this case.

This is a latent defect present on current `main`, not a recent regression — it surfaced as the CLIs grew slower and Gemini crossed the line. It is independent of the other open PRs: none of them touches `controlTimeoutMs` or the client startup path.

Same class as #146 (a window mistaken for a budget) and #137 (an error collapsed, its information lost), but different code and a different defect.

## Changes

- `AcpStartupError` — new.
- `AcpContentPayload` — refusal origin `startup`.
- `describeAcpSessionOpenFailure` — a startup branch that does not talk about sessions.
- `ManagedAcpExecutionBackend` — `startupTimeoutMs`, `AcpStartupError`, and carrying the refusal.
- The six ACP compositions — pass the refusal origin through.

## Verification

- Three new tests, each watched failing on the base before the change (`3 failed, 46 passed`).
- Named delta of failing tests across all ACP providers against `main`: empty in both directions.
- Full suite: 18 suites / 30 tests failed, 9155 passed — identical to this machine's Windows background.
- `typecheck` and `eslint` clean over the touched directories.
- Verified live after install: all three affected providers started conversations and answered; Grok, Claude and Codex unaffected.

## Not established

The default of 30 s was chosen as comfortably sufficient rather than measured against the slowest cold start.


greenkms:
I apologize for the large amount of PR, but the new version required the elimination of many bugs

sandsaber:
@greenkms no worries! Big thanks for hard work! Really appreciated! Cannot catch everything :(

sandsaber:
@greenkms you are outstanding! Thank you, that you are putting significant effort in this project!

greenkms:
Rebuilt on top of current `main` (`e6f6de45`), since #146 landed there in the meantime.

The two branches collided in one place only: `FakeScheduler` in
`tests/unit/providers/opencode/OpencodeExecutionBackend.test.ts`. Both taught the
fake scheduler to remember a timer's delay, independently — #146 through a parallel
`delays` map feeding `pending(ms)`, this branch through a `{run, delayMs}` record
feeding `fireAllUpTo`.

Resolved in favour of `main`'s storage: `tasks` + `delays` and `pending()` are kept
exactly as merged, and `fireAllUpTo` is reimplemented on top of them. `fireNext` and
`fireLast` now also drop the handle from `delays`, which `main` omitted — without it
a fired timer stayed visible to `pending()`.

No production code was touched by the merge. `typecheck`, `lint`, `review:css` and
`build:release` are green; the unit suite shows the same 22 pre-existing Windows
failures by name as `e6f6de45` itself, with an empty delta in both directions.

PR #155, opened 2026-09-07
Nordic: the context manager, and three defects only a render could find

sandsaber:
The tail of the Nordic redesign: the chat surfaces the drawing still had open, the
context manager made reachable, and three defects that only a render could find.

Not for merge yet — 2.0.0 is unreleased and this wants a pass in a real vault first.

## Context management

The dialog whose one job is deciding what to keep could not remove a file. Obsidian's
`.modal` is 560px wide with 16px of padding; the dialog set its 640 on the *content*, so
the host clipped the right column off — every row's token count, every row's remove cross,
Add, and Done. The harness reported both numbers before anything was changed, which is the
whole diagnosis.

With that fixed, the rest followed:

- **It is reachable.** Manage lived on the summary row, and the summary row only existed
  past four attachments — so with two files attached there was no way in at all. Now: a
  list control pinned in the chip row's corner, out of the row's flow so a narrow sidebar
  cannot push it to a line of its own; and a `manage-context` command a shortcut can bind.
- **The threshold is one.** Four chips wrap to three lines at every width the plugin is
  actually used at, and the second row echoing three of five names carried no remove on
  them — something to read and nothing to do. One chip always fits; past it a line that
  never wraps, the same height as the chip row it replaces, so attaching a file no longer
  shrinks the box you write in.
- **The picker can reach outside the vault.** A file and a folder are asked for
  separately: Windows and Linux cannot show one dialog that returns either, and Electron
  answers a request for both with the folder picker — so one "browse" control is a control
  that cannot attach a file there. Opened from the dialog it mounts *inside* it, because a
  modal is its own stacking context and a panel drawn in the composer lands behind it at
  any z-index.
- **One close.** The host draws its own over that corner and the header drew a second
  under it. Hiding the host's meant knowing where and when it builds the thing — through
  `.modal` it is not there, at `onOpen` it does not exist yet. Drawing none needs neither
  answer.

## A class alone loses to the host

Obsidian names elements and attributes; Grimoire names classes. Three ways that bit, each
found in a render rather than in a reading:

- `.grimoire-icon-btn--small` never reached its own width and height, because
  `button.grimoire-icon-btn` outweighs it — every small icon button drew four pixels large.
  `designSystem.test.ts` now fails any `.x--m` that sets a property an `element.x` rule
  also sets; written first, it found exactly two violations and nothing else.
- A focused field kept the host's 2px ring from `input[type='text']:focus-visible`, which
  for a panel that focuses its field as it opens is permanent chrome around a band both
  3b and 3c draw flat. Focus reads on the glyph now.
- A modal is sized on `.modal`, not on what it holds.

The gates cannot see a rule the plugin never wrote — a missing `:focus` rule violates
nothing — so §16 of the design system now says the render is the check that finds these.

## Also in here

- A saved turn no longer stores its conversation twice. The tab's binding lives in two
  places; a save that trusted only `state.currentConversationId` found it null at the end
  of a turn, took the runtime's live session for an id, and wrote ten messages again under
  it. `isSessionClaimed` answers what the store cannot see.
- A stored transcript is repaired as it is read: nothing it calls running is running, so a
  conversation stored mid-turn reopened with a spinner that could never stop.
- Grok's four commonest tools were missing from normalization entirely and drew as a
  wrench beside a snake_case word. Codex's `request_user_input` has no item in its item
  stream, so an answered question kept its spinner until the turn ended.
- `blocked` and `unfinished` stop looking identical. One is a decision; the other is the
  absence of one, and drawing both in the error colour told the reader they had refused
  something they never saw.

## Checks

`npm run typecheck`, `npm run lint`, `npm run review:css` and 9121 unit tests pass. Every
layout claim above was measured in a headless render carrying Obsidian 1.13.7's own
`app.css`, before and after. Every new gate was watched to fail first.

## Also

`docs/design/Grimoire Nordic.html` — the drawing the design system is *of* — is now checked
in. Three docs link to it and §16 tells the reader to open it and compare a render against
it; untracked, that instruction was dead for everyone but the author. `AGENTS.md` names the
directory as the exception to the handoff rule, since the drawing moves with the design
rather than being cleaned up after it.

## Known, deliberately left

- At 2–4 files the composer no longer shows names, only a count. Cheapest reversal if it
  grates: raise `CHIP_THRESHOLD`, without bringing the echo row back.
- `common.close` is now unused in ten locales.


sandsaber:
@greenkms you are welcome to look and test :)

greenkms:
OMG so many new things.
I hope there won't be many bugs.

greenkms:
Interesting new design, some parts are pleasing, for example, moving around the topic has become much more convenient.

PR #162, opened 2026-09-08
feat(title): record and show where a title came from

### LINKED ISSUE #161: Nothing records who wrote a conversation's title, so the UI reads the wrong field
## Summary

There is no record of where a conversation's current title came from. The closest thing is `titleGenerationStatus`, and it answers a different question: what the last *attempt* at generating did. The UI reads it as though it described the title, and the two diverge in both directions.

**A hand-written title keeps looking like a failure.** `renameConversation` writes the title and nothing else:

```ts
await this.storage.sessions.updateMetadata(conversation, ['title']);
```

so a conversation the user renames after a failed generation keeps `titleGenerationStatus: "failed"` for ever. The history row goes on showing the `refresh-cw` regenerate control for a name that is the user's own. On the vault this was found on, 21 rows out of 199 are in that state.

**An absent status says two different things.** A conversation that never generated a title and one the user renamed while it was generating — the branch that clears the status to `undefined` — are indistinguishable afterwards.

**The placeholder is not recorded at all.** Whether the title now shown is the fallback cut from the first message or something a model wrote can only be guessed at by comparing the title against the message text. Doing exactly that over 232 conversations: 152 of the 173 `failed` records look like placeholders, 26 `success` records look like none, and the remaining 21 are the hand-renamed ones above.

## Consequences

- «Regenerate title» in the history context menu is gated on `titleGenerationStatus === 'failed'`, so a title the model wrote but the user dislikes cannot be regenerated from the history at all — only from the tab menu. A title one is not happy with is not a failed one.
- The regenerate icon marks the wrong rows: present on hand-renamed conversations, absent on placeholders whose last attempt happened to succeed earlier.
- Nothing in the UI distinguishes a generated title from a placeholder, which is the distinction a user scanning a tab list actually wants.

## Expected

The record says who wrote the title being shown — the placeholder, the model, or the user — and the UI reads that rather than inferring it from the status of an attempt. Conversations written before such a field exists should say nothing rather than be guessed at.

greenkms:
<html>
<body>
<h3 data-heading="Problem">Problem</h3>
<p><code>titleGenerationStatus</code> describes the last <em>attempt</em> at generating a title. The<br>
UI reads it as though it described the title, and the two are not the same<br>
thing.</p>
<p><code>renameConversation</code> writes the title and nothing else, so a conversation the<br>
user renames by hand after a failed generation keeps <code>failed</code> for ever — and the<br>
history row goes on offering to regenerate a name that is the user's own. On the<br>
vault this was found on that is 21 rows out of 199. In the other direction an<br>
absent status cannot distinguish a conversation that never generated a title<br>
from one the user renamed while it was generating.</p>
<p>There is also no way to ask the question directly: whether the title now shown<br>
is the placeholder cut from the first message or something a model wrote is<br>
recoverable only by comparing the title against the message text.</p>
<h3 data-heading="What changed">What changed</h3>
<p><code>titleSource</code> — <code>'fallback' | 'model' | 'manual'</code> — records who wrote the title<br>
being shown, beside the status that records what the last attempt did.<br>
<code>renameConversation</code> takes it as an argument and writes it together with the<br>
title, so the two cannot disagree. A caller with nothing to say about<br>
provenance — a fork, a duplicate, a restore — omits it and leaves the recorded<br>
source alone rather than blanking it.</p>
<p>The field is optional. Conversations written before it say nothing rather than<br>
guessing, and every reader falls back to the previous behaviour for them.</p>
<p>Shown as a coloured star beside each rendering of the name:</p>

Where | What
-- | --
Tab context-menu heading | star before the title
History row | star before the title
Rename dialog | star beside the field, never inside the value
Tab tooltip | the word for a placeholder, nothing otherwise

<p>The tab badge is deliberately not a caller: it shows a number, not a name. The<br>
rename input is not one either — a prefix inside the edited value would be saved<br>
as part of the title. The tab's tooltip is an <code>aria-label</code>, a string with no DOM<br>
to colour and a reader who cannot see colour, so it names the placeholder in<br>
words instead of showing the star.</p>
<p>Two behaviours follow the new field rather than the status: the regenerate icon<br>
in the history row marks a placeholder rather than a failed attempt, and<br>
«Regenerate title» in the history context menu is offered whatever the title is,<br>
gated only by <code>canSuggestTitle</code>. A title one is not happy with is not a failed<br>
one, and regenerating a good title was previously reachable only from the tab<br>
menu.</p>
<h3 data-heading="Tests">Tests</h3>
<p>Added: the source travels with the title through the metadata writer and is not<br>
blanked by a writer that omits it; <code>getTabTitleSource</code> prefers a tab-local name<br>
over the conversation's record; the tooltip announces a placeholder and stays<br>
silent otherwise; the menu heading marks a short title and marks a long one<br>
without losing the shortening; the history row marks a known source and says<br>
nothing for a conversation older than the field; the regenerate control follows<br>
the placeholder rather than the last attempt; the rename dialog puts the marker<br>
beside the field and leaves the input value clean.</p>
<p>Updated: the assertions that pinned <code>renameConversation</code>'s argument list, one<br>
history-menu item list that now includes the unconditional regenerate entry, and<br>
two entries in the locale-coverage allowlist for <code>Manual</code>, which is spelled the<br>
same in Spanish and Portuguese as in English.</p>
<h3 data-heading="Rebased onto the merged design system">Rebased onto the merged design system</h3>
<p>The branch is <code>feat/title-source-marker-v2</code>, rebased onto <code>main</code> after #152 and<br>
#155 landed. Three resolutions were needed and none of them is mechanical:</p>
<ul>
<li><code>TabBar.getAccessibleTitle</code> now composes <code>chat.ui.tabs.tabLabel</code> from the index<br>
and the title. The placeholder word goes into the title half, so the label<br>
reads <code>Tab 2 · Placeholder · My Conversation</code> rather than replacing the new<br>
format.</li>
<li>The rename dialog dropped its label — the dialog's own heading names the field<br>
now — so the marker attaches to the field container instead, still beside the<br>
input and still never inside the value.</li>
<li>The history row's title is now <code>conv.title?.trim() || formatDate(...)</code> with the<br>
meta folded into the <code>title</code> attribute. The star was added to that, not in<br>
place of it.</li>
</ul>
<p><code>--text-muted</code> and <code>--text-success</code> were replaced with <code>--grimoire-ink-muted</code> and<br>
<code>--grimoire-ok</code>, and <code>--grimoire-space-1</code> with <code>--grimoire-space-2</code>, as the new<br>
<code>designSystem</code> test requires: Obsidian variables belong to the token layer, and<br>
a token a stylesheet reads must be defined.</p>
<h3 data-heading="Verification">Verification</h3>
<p><code>typecheck</code>, <code>lint</code>, <code>review:source</code>, <code>review:css</code>, <code>review:deps</code>,<br>
<code>build:release</code> — all pass.</p>
<p><code>npx jest --selectProjects unit</code> — the Windows baseline of 22 failing test names<br>
is unchanged. Compared by name against <code>origin/main</code> with <code>comm</code>: the delta is<br>
empty in both directions.</p>
</body>
</html>

Fixes #161


greenkms:
Thank you for persistently raking my PR

PR #165, opened 2026-09-08
fix(chat): say which model answered without hovering the history row

### LINKED ISSUE #164: The history list no longer says which agent answered, and a tooltip cannot be scanned, tabbed to, or tapped
## Summary

Since the design pass, a row in the history list is the title and a timestamp. The model that answered — along with the prompt preview, source count and usage — moved into the `title` attribute of the title element, so it is reachable only by hovering one row at a time.

Three of those four facts belong in a tooltip. They describe a conversation you have already picked out. The model is not like them: it is what tells two rows apart *before* you open either. With a list that reads `Chapter 03.md`, `Fix the build`, `Chapter 03.md`, the title alone does not say which one was Codex and which was Claude.

## Why a tooltip cannot carry this one

- **It cannot be scanned.** Finding a conversation means running your eye down the list. A tooltip shows one row's worth at a time, so a list of twenty is twenty pointer movements and twenty waits.
- **It never opens for a keyboard user.** The row is focusable — `tabindex` is set on it deliberately — and `title` does not surface on focus. Tab through the list and the model is not there at all.
- **On touch it does not exist.** There is no hover on a tablet.

The row already commits to being scannable: the title is clipped with an ellipsis rather than wrapped, and the stamp yields to it. Losing the model is what makes scanning stop working — the reader gets to the right title and still has to point at it to learn whether it was the agent they meant.

## What the row looked like before

`ConversationController.renderHistoryItem` built a second line:

```ts
content.createDiv({
  cls: 'grimoire-history-item-meta',
  text: this.formatHistoryMeta(conv),   // model · preview · sources · usage
});
```

and `.grimoire-history-item-meta` in `history.css` set it faint and clipped. Both are gone; `formatHistoryMeta` survives, feeding the tooltip only.

The comment left in place states the reasoning, and I think it is right about what it names:

> The row carried a second line of provider, prompt preview, source count and usage - four facts in 11px mono under every title, which turned a list you scan into a wall you read.

Four facts under every title is a wall. The disagreement is only about whether all four had to go: the preview is a sentence, the counts are numbers you look up, and those are what made the line heavy. The model is two or three words and is the reason the line existed.

## Proposed fix

PR — the model alone comes back as its own line, faint, one step below the title, clipped rather than wrapped so a billing-path label like `Z.AI Coding Plan/GLM-5.3-Flash` cannot grow the row. Preview, sources and usage stay in the tooltip. A conversation that never recorded a model draws no line at all, so untitled rows do not gain an empty one.

That keeps the density argument intact — one fact, not four — while putting the identifying fact back where it can be read without a pointer.

If a second line is unwanted at any price, the alternative is to give the model its own place in the existing line rather than under it. I did not take that route: the row's right edge is where `.grimoire-history-item-actions` appears on hover, and `.grimoire-history-item-time` is deliberately faded out to make room for it, so anything parked there would disappear exactly when the reader is aiming at the row.

## Environment

Grimoire 2.0.0 at `e6f6de45`, Obsidian 1.13.7, Windows 11. Observed in a vault where the history list runs to hundreds of conversations across six providers, which is where scanning matters and where the loss is noticeable.

greenkms:
## What this changes

The history row gets one line back: the model that answered, under the title.

The design pass moved four facts out of the row and into its tooltip — provider, prompt preview, source count and usage. Three of them answer a question about a conversation you have already found. The model answers the question you ask *before* you open anything, and a tooltip cannot answer it: it cannot be scanned down a list, it does not surface on keyboard focus even though the row is focusable, and on touch there is no hover at all.

So the model alone returns, and the other three stay where the design pass put them.

## The line

```ts
const modelLabel = conv.modelLabel?.trim();
if (modelLabel) {
  content.createDiv({ cls: 'grimoire-history-item-model', text: modelLabel });
}
```

- **Faint and one step down** (`--grimoire-ink-faint`, `--grimoire-text-2xs`), so the row still reads title first.
- **Clipped, never wrapped.** A label can arrive as a whole billing path — `Z.AI Coding Plan/GLM-5.3-Flash` — and a second line that grows is exactly the wall the single line was removed to avoid. `text-overflow: ellipsis` with `white-space: nowrap`, same treatment the title already has.
- **Absent when there is nothing to say.** A conversation that never recorded a model draws no element, so a row does not gain an empty line in whitespace.

No new tokens, no literal sizes, no Obsidian variable read directly; the `designSystem` gate passes unchanged.

## Tests

Two, both written before the change and watched to fail:

- `says which model answered without being hovered, and keeps the rest in the tooltip` — the model is text in the row; preview, sources and usage are not, and are still in the tooltip. This is the one that failed before the fix.
- `draws no model line for a conversation that never recorded one`.

The existing `keeps model, sources and usage as the row's tooltip, not a second line` is left as it is and still passes: it asserts on `.grimoire-history-item-meta`, the four-fact line, which does not come back. The new element is a different class carrying one fact.

## Checks

`npm run typecheck`, `npm run lint`, `npm run review:css` clean.

`npx jest --selectProjects unit`: 22 pre-existing failures on this Windows host, identical by name to `e6f6de45` itself — empty delta in both directions. `ConversationController`, `designSystem` and the history suites: 348 passed.

## What I am not proposing

Not a revert of the density change: the second line is one fact now rather than four. If the row must stay single-line at any price, the model would have to go into the existing line instead — but not at its right edge, where `.grimoire-history-item-actions` takes over on hover and `.grimoire-history-item-time` is faded out to make room, which would hide the label exactly when the reader is aiming at the row.

Happy to put it behind a setting instead, the way `showPanelLabels` is, if you would rather the default row stayed as drawn.

Fixes #164


greenkms:
The last of the PR pile. So far everything looks stable.

sandsaber:
@greenkms you rock!

greenkms:
I actually missed the bug with the double names. I'm sorry.

PR #185, opened 2026-09-10
fix(kimicode): detect default Windows executable

### LINKED ISSUE #184: Kimi Code default Windows executable is not auto-detected
## Summary

Grimoire does not auto-detect Kimi Code CLI at its default Windows installation
path. The provider works after manually entering the executable path in settings.

## Environment

- Windows desktop
- Kimi Code CLI 0.42.0
- Default executable: `C:\Users\<user>\.kimi-code\bin\kimi.exe`

## Steps To Reproduce

1. Install Kimi Code CLI in its default Windows location.
2. Restart Obsidian so the provider performs CLI discovery.
3. Enable Kimi Code without setting a custom CLI path.

## Actual

Kimi Code is not detected automatically. Entering the full path to `kimi.exe`
in **Kimi Code > CLI path** makes the provider available.

## Expected

Grimoire should detect `~/.kimi-code/bin/kimi.exe` automatically on Windows,
without requiring a host-specific manual path.

## Root Cause

`KimicodeCliResolver` checks the known-install fallback
`~/.kimi-code/bin/kimi`. The shared resolver applies `PATHEXT` while searching
directories from `PATH`, but treats explicit fallback paths as complete file
names. It therefore never probes the actual Windows file, `kimi.exe`.

## Workaround

Set the provider CLI path manually to the full `kimi.exe` path.

greenkms:
## Problem

Kimi Code installs its Windows launcher at `~/.kimi-code/bin/kimi.exe`, but
Grimoire's known-install fallback only probes the extensionless
`~/.kimi-code/bin/kimi`. Users must configure the executable manually when the
Obsidian process cannot discover the newly installed CLI through `PATH`.

Fixes #184 

## Root Cause

`resolveCliExecutable` expands `PATHEXT` candidates when searching directories
from `PATH`. Explicit fallback paths are validated as complete file paths, so
the extensionless Kimi fallback never resolves the real Windows executable.

## Approach

- Add `~/.kimi-code/bin/kimi.exe` as the first Kimi-owned known-install fallback.
- Preserve the existing extensionless fallback for other platforms.
- Add a focused regression test for the default Windows installation path.

The change remains provider-local and does not alter shared CLI resolution for
other providers.

## Verification

Automated:

- Focused `KimicodeCliResolver` unit suite: 4 passed.
- `npm run typecheck`: passed.
- `npm run lint`: passed.
- `npm run build:release`: passed, including source/CSS/dependency review,
  release metadata validation, isolated bundle load, and `grimoire-view` smoke check.
- The full unit suite retains two unrelated host-locale failures on Windows
  (`16,000` versus `16 000`); the affected files are unchanged from `origin/main`.

Manual:

- Built commit `eec473d2` and installed it in Windows desktop Obsidian.
- Removed the manually configured Kimi Code CLI path after restarting Obsidian.
- Grimoire then detected `~/.kimi-code/bin/kimi.exe` automatically and the Kimi
  provider remained connected.

## Security And Privacy

- No credentials, account data, prompts, or logs are read or persisted.
- Process arguments and provider permissions are unchanged.
- The change only adds one deterministic local executable fallback candidate.

### DIFF
--- a/src/providers/kimicode/runtime/KimicodeCliResolver.ts
+++ b/src/providers/kimicode/runtime/KimicodeCliResolver.ts
@@ -7,6 +7,7 @@ import { resolveCliExecutable } from '../../../utils/resolveCliExecutable';
 import { getKimicodeProviderSettings } from '../settings';
 
 const KIMI_CODE_DEFAULT_BIN = path.join(os.homedir(), '.kimi-code', 'bin', 'kimi');
+const KIMI_CODE_DEFAULT_WINDOWS_BIN = `${KIMI_CODE_DEFAULT_BIN}.exe`;
 
 export class KimicodeCliResolver {
   private readonly cachedHostname = getHostnameKey();
@@ -48,7 +49,7 @@ export class KimicodeCliResolver {
   ): string | null {
     const hostnamePath = (hostnamePaths?.[this.cachedHostname] ?? '').trim();
     return resolveCliExecutable('kimi', [hostnamePath, legacyPath], envText, {
-      fallbackPaths: [KIMI_CODE_DEFAULT_BIN],
+      fallbackPaths: [KIMI_CODE_DEFAULT_WINDOWS_BIN, KIMI_CODE_DEFAULT_BIN],
     });
   }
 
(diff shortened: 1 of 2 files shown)


greenkms:
Superseded by #187, which includes this Kimi Code fix together with the confirmed Windows auto-detection fixes for MiMoCode, Codex, and Grok. Closing this PR to avoid overlapping changes.

